PR #32, opened 2016-04-10
Show PE entrypoint

dmex:
This patch shows the AddressOfEntryPoint field from the OptionalHeader of any PE binary when viewed with peview.

There are two feature requests for this feature:
https://wj32.org/processhacker/forums/viewtopic.php?f=14&t=1803
https://wj32.org/processhacker/forums/viewtopic.php?f=14&t=2313&p=7472


dmex:
Pull request has errored...

PR #37, opened 2016-04-18
Updater UI rewrite

dmex:
Completed initial updater rewrite


wj32:
Have you tested this with error conditions (e.g. no network connection, hash failed, etc.)?

dmex:
Yeah, there are defines in updater.h for testing errors but not for failed network connections... It shows a generic error page and might need some special handling.

wj32:
What's the point of InitializedEvent? You don't seem to wait on it anywhere.

dmex:
ShowUpdateDialog waits on the InitializedEvent since it needs a valid window handle to send the WM_SHOWDIALOG notification

PR #38, opened 2016-04-28
Updater signing

dmex:
Adds an extra layer of verification for downloads by the updater.

Four things need to be resolved:

- [x] UpdaterTrustedPublicKey in verify.c needs to be changed (We could use the same keypair as KPH but it really should be a separate keypair?).
- [x] processhacker-2.39-setup.exe needs to be signed with the public key that we're using in verify.c.
- [x] CustomSignTool.exe needs to pass the verify signature buffer into PhBufferToHexString and save the hex string into a file instead of just saving the buffer - otherwise we can't transmit the signature via XML in update.php.
- [x] Update config.php on the server with the hex result from the signature.


wj32:
If we're going to do this then we don't need to call PhVerifyFile to check yet another signature...

dmex:
Removed PhVerifyFile

PR #45, opened 2016-06-09
installer: Change "Process Hacker 2" to "Process Hacker".

dmex:
Initial pull request for installer branch.

### DIFF
--- a/build/installer/Process_Hacker_installer.iss
+++ b/build/installer/Process_Hacker_installer.iss
@@ -60,8 +60,8 @@ AppPublisherURL={#website_url}
 AppSupportURL=http://wj32.org/processhacker/forums/
 AppUpdatesURL={#website_url}
 UninstallDisplayName=Process Hacker {#app_version_full}
-DefaultDirName={pf}\Process Hacker 2
-DefaultGroupName=Process Hacker 2
+DefaultDirName={pf}\Process Hacker
+DefaultGroupName=Process Hacker
 VersionInfoCompany=wj32
 VersionInfoCopyright={#copyright}
 VersionInfoDescription=Process Hacker Setup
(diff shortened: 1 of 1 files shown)


dmex:
Are you able to rename the shortcuts, add/remove programs entries and other strings from "Process Hacker 2" to just "Process Hacker"?

XhmikosR:
The problem is that we'll end up with duplicate shortcuts unless I remove the old ones.

Which ones are you talking about **exactly**?

dmex:
Desktop and Startmenu shortcuts, name shown for Programs and Features (ARP) entries and the default install folder path (the default path used when there's no existing installation)?

XhmikosR:
If it's an upgrade, how do you want to proceed?

XhmikosR:
Also, how about `AppUserModelID`? Should I rename that too?

XhmikosR:
Also, the appdata folder is still Process Hacker 2.

/CC @wj32 for your overall thoughts

dmex:
Yeah, rename as much as possible from "Process Hacker 2" to just "Process Hacker".

dmex:
@XhmikosR Is this alright to merge?

XhmikosR:
@dmex: nope and I can't build the latest master on my dev VM with Windows 7 so I can't test.

dmex:
@XhmikosR 

The latest master builds fine on my win7 machine?

XhmikosR:
I get numerous errors when makesdk runs. Where am I supposed to call that
from?

dmex:
@XhmikosR 

Have you tried makesdk_rebuild.cmd to build the release binaries (at least once) or does that also error?

XhmikosR:
I'll check it out when I'm back. IIRC I don't call that in my script.

XhmikosR:
I'm getting the same errors with makesdk_rebuild.cmd.

```
Set-Location : Cannot process argument because the value of argument "path" is null. Change the value of argument "path" to a non-null value.
At C:\Users\xmr\Desktop\processhacker2\build\sdk\makesdk.ps1:44 char:21
+         Set-Location <<<<  $PSScriptRoot;
    + CategoryInfo          : InvalidArgument: (:) [Set-Location], PSArgumentNullException
    + FullyQualifiedErrorId : ArgumentNull,Microsoft.PowerShell.Commands.SetLocationCommand
```

dmex:
@XhmikosR 

Can you open powershell and execute Get-Host (or type $PSVersionTable and hit enter) and paste the version number?

XhmikosR:
```
Windows PowerShell
Copyright (C) 2009 Microsoft Corporation. All rights reserved.

PS C:\Users\xmr> Get-Host

Name             : ConsoleHost
Version          : 2.0
CurrentCulture   : el-GR
CurrentUICulture : en-US
PrivateData      : Microsoft.PowerShell.ConsoleHost+ConsoleColorProxy
IsRunspacePushed : False
Runspace         : System.Management.Automation.Runspaces.LocalRunspace
```

dmex:
You're running the Windows 7 RTM version of Powershell from 2009 and that version doesn't support the functionality used by the build script... I'll add a version check to the build script so it doesn't error.

You need to install PowerShell 4.0 or later. The latest version is 5.1 available here: https://www.microsoft.com/en-us/download/details.aspx?id=54616

Extract Win7AndW2K8R2-KB3191566-x64.zip and run the installer.

XhmikosR:
So much crap needed for so simple tasks. You should definitely simplify the requirements. The same goes for the SDK.

dmex:
There's only one requirement: 'keeping your machine updated' 😄 

Powershell is needed for multiple reasons and it's easier to build PH than it was previously...

* makesdk_rebuild.sdk for building _everything_ needed by the sdk and the sdk itself.
* makesdk.cmd for just updating the sdk files (if you have already compiled Process Hacker using Visual Studio).

In the past you had to open the main ProcessHacker.sln solution, build Process Hacker, change the solution configuration and platform, build PH again etc... 4 times before you could even run makesdk.sdk and it just wasn't intuitive for plugin developers and people were building things incorrectly then complaining about it :/

The build system also needed to support multiple use cases:
* New developers
* Plugin developers
* Nightly builds
* Appveyors cloud based build system
* Encryption and decryption
* Compiler debugging (and configurable debug output)
* Updating the website with the latest build information (e.g. https://wj32.org/processhacker/nightly.php)
* Future-proof (cmd.exe is deprecated and doesn't support any of the major console improvements on Win10. Powershell is also now the default interpreter starting with the Anniversary update coming in April)

Simple bat scripts just don't cut it for these workloads and nothing has really changed, you only need to run makesdk_rebuild.sdk if you want to build the sdk (and keep your machine updated) 😃

XhmikosR:
Either way, I haven't been following the development and it appears to me you have made other changes in the installer. Since I can't go now and review any old changes due to lack of time, I'm gonna close this and feel free to add you solution.

PR #53, opened 2016-08-03
Tree: Return to tree mode when clicking Name...

orgads:
... after sorting by another column.


dmex:
Is this needed when you can right-click the name column and select Reset to return to tree mode?

http://i.imgur.com/b6pekna.png

orgads:
I wasn't aware of that feature. It requires more clicks, but ok :)

PR #57, opened 2016-08-19
Add highlighting for Pico processes (Windows subsystem for linux)

dmex:
Add highlighting for Pico processes (Windows subsystem for linux)


dmex:
This is also required for later changes that are needed to support Pico processes in Process Hacker.

PR #123, opened 2017-04-18
[peview] Add safeguards around using loadconfig most recent members

lucasg:
Since the last PR on `CFG` tab for `peview` (https://github.com/processhacker2/processhacker2/pull/101), I've noticed some `PE` has a `GuardFunctionTable` but does not define a `GuardAddressTakenIatEntryTable` or a `GuardLongJumpTargetTable` (mainly `Wow64` ones). 

This PR add safeguards when parsing the load configuration, in order to read only what's defined and no more. This is basically the `NT` loader 's parsing algorithm for load configurations.


dmex:
LGTM 👍

PR #124, opened 2017-04-18
[WIP] Heuristic to locate and tag the CFG Bitmap in memory view

lucasg:
This PR is a proposal to tag the CFG bitmap memory allocation in the "Memory" view of a process. It might help puzzled developers to understand why there is a giant 2 TB  memory allocation in their process memory map.

64-bit process which are CFG-aware allocate a 2 TB reserved memory
map to store the CFG bitmap used to determinate the valid target call.

This bitmap base address cannot - yet - be located using public API (that
I know of) therefore we need to rely on heuristics.

Before : 
cfg_bitmap_naked

After:
cfg_bitmap_tagged

### DIFF
--- a/ProcessHacker/include/memprv.h
+++ b/ProcessHacker/include/memprv.h
@@ -20,7 +20,8 @@ typedef enum _PH_MEMORY_REGION_TYPE
     HeapRegion,
     Heap32Region,
     HeapSegmentRegion,
-    HeapSegment32Region
+    HeapSegment32Region,
+    CfgBitmapRegion
 } PH_MEMORY_REGION_TYPE;
 
 typedef struct _PH_MEMORY_ITEM
--- a/ProcessHacker/memlist.c
+++ b/ProcessHacker/memlist.c
@@ -424,6 +424,8 @@ PPH_STRING PhpGetMemoryRegionUseText(
     case HeapSegment32Region:
         return PhFormatString(L"Heap segment%s (ID %u)",
             type == HeapSegment32Region ? L" 32-bit" : L"", (ULONG)MemoryItem->u.HeapSegment.HeapItem->u.Heap.Index + 1);
+    case CfgBitmapRegion:
+        return PhFormatString(L"CFG Bitmap");
     default:
         return PhReferenceEmptyString();
     }
--- a/ProcessHacker/memprv.c
+++ b/ProcessHacker/memprv.c
@@ -406,58 +406,68 @@ NTSTATUS PhpUpdateMemoryRegionTypes(
     // Locate CFG Bitmap for 64-bit process
     if (!isWow64)
     {
-        BOOL bFoundCfgBitmap = FALSE;
-        size_t CurrentAllocationSize = 0;
-        PVOID CurrentAllocationBase = NULL;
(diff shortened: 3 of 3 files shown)


ionescu007:
Hey Lucas,

A heuristic is not needed. The address of both bitmaps (WoW32 + x64) is in
the export of Ntdll.dll called LdrStartupInitBlock. Please read it from
there.

Best regards,
Alex Ionescu

lucasg:
You mean `LdrSystemDllInitBlock` right ? That's probably a better solution.

One thing though, I haven't found a type or a structure to parse this block. It's a `size+payload` type of data structure (like the load configuration) but neither `IDA` nor `windbg` gives me a proper structure for it.

In that case we  have to rely on hardcoded offsets.

dmex:
@lucasg 

Try this code and let me know if it works:

```
typedef struct _LDR_INIT_BLOCK
{
    ULONG cbSize;
    PVOID Unknown1;
    PVOID CfgBitmapAddress;
    PVOID Unknown2[23];
} LDR_INIT_BLOCK, *PLDR_INIT_BLOCK;

static PLDR_INIT_BLOCK (WINAPI *LdrSystemDllInitBlock_I)(
    VOID
    );

PLDR_INIT_BLOCK ldrInitBlock = NULL;
PPH_MEMORY_ITEM cfgBitmapMemoryItem;
PPH_STRING ntdllFileName = PhConcatStrings2(USER_SHARED_DATA->NtSystemRoot, L"\\System32\\ntdll.dll");

status = PhGetProcedureAddressRemote(
    ProcessHandle,
    ntdllFileName->Buffer,
    "LdrSystemDllInitBlock",
    0,
    &ldrInitBlock,
    NULL
    );
PhDereferenceObject(ntdllFileName);

if (!NT_SUCCESS(status))
    return status;

if (ldrInitBlock->cbSize == sizeof(LDR_INIT_BLOCK))
{
    if (cfgBitmapMemoryItem = PhLookupMemoryItemList(List, ldrInitBlock->CfgBitmapAddress))
    {
        memoryItem->RegionType = CfgBitmapRegion;
    }
}
```

lucasg:
Your code does the job, minus some modifications:

1. I don't have the same `LDR_INIT_BLOCK` structure. (Windows10, build 15063)
```
typedef struct _LDR_INIT_BLOCK
{
	ULONG cbSize;
	PVOID Unknown1[21];
	PVOID CfgBitmapAddress;
	PVOID Unknown2[3];
} LDR_INIT_BLOCK, *PLDR_INIT_BLOCK;
```

2. We actually need to read the target process's memory to retrieve the init block.

I've updated the PR.

dmex:
I've tidied up the code and merged the PR 👍 

I didn't have a compiler/debugger handy when I created that typedef, surprised it worked at all.

ionescu007:
The other address is the Wow64 bitmap. Remember that a 32-bit application
in x64 Windows will have both. So this needs to be fixed to handle both
bitmaps.

Best regards,
Alex Ionescu

lucasg:
Good catch, I didn't thought of wow64 process. Anyway, I will get on it once I have some free time.

PR #125, opened 2017-04-20
[pe] List exports by ordinal as well as by name

lucasg:
`PeViewer` does not list exports by ordinal. 

`ClangDll.def`
``` 
LIBRARY

EXPORTS  
   fun1		@1  NONAME 
   fun2		@2
   fun3		@3  NONAME   
   fun4		@4
   data1	@5
   data2	@6  NONAME
```

[ClangDll.zip](https://github.com/processhacker2/processhacker2/files/942360/ClangDll.zip)

Results :

* Ida : 
ida_export_parsing

* Depends :
depends_export_parsing

* Peviewer:
peview_master_export_parsing

The "bug" was considering `Exports->OrdinalTable` as the ordinal table for all exports, whereas it's only for **named exports**.

Peviewer with "fixed" exports listing :
peview_pe_export_parsing

NB : there is probably a bug with how the VA are computed, since it's not entirerly consistent with `ida` or `depends`.

### DIFF
--- a/phlib/mapimg.c
+++ b/phlib/mapimg.c
@@ -624,8 +624,8 @@ NTSTATUS PhGetMappedImageExports(
             );
         PhpMappedImageProbe(
             MappedImage,
-            Exports->OrdinalTable,
-            exportDirectory->NumberOfFunctions * sizeof(USHORT)
+            Exports->OrdinalTable,  // ordinal list for named exports
+            exportDirectory->NumberOfNames * sizeof(USHORT)
             );
     }
     __except (EXCEPTION_EXECUTE_HANDLER)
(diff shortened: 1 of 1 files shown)


dmex:
Looks good, thanks!

dmex:
>there is probably a bug with how the VA are computed

user32.dll is a good example of unnamed exports with invalid VA:
http://i.imgur.com/3IyrUN0.png

lucasg:
I've looked into it yesterday and here's my guess : unlike most PE parsers (`dumpbin.exe`, `pefile`, etc.) which works on the file itself, `phlib` map the PE in memory. I don't know why it has an impact on the computed VA, but it does.

If you look at [`radare2`, it compute the exports address the following way](https://github.com/radare/radare2/blob/7327642241a5a2fb3c278e86799a663a5013c217/libr/bin/format/pe/pe.c#L2414) :

* `vaddr` (virtual address)  by rebasing against the PE's base address : https://github.com/radare/radare2/blob/7327642241a5a2fb3c278e86799a663a5013c217/libr/bin/format/pe/pe.c#L306
* `paddr` (physical address, as in file offset ?) by translating the `rva` into `va` : https://github.com/radare/radare2/blob/7327642241a5a2fb3c278e86799a663a5013c217/libr/bin/format/pe/pe.c#L278

It makes no sense, since the official PE-COFF spec sheet explicitely say it's a `rva`.

Anyway here's the patch :

    diff --git a/phlib/mapimg.c b/phlib/mapimg.c
    index 05d179f..eae0ce4 100644
    --- a/phlib/mapimg.c
    +++ b/phlib/mapimg.c
    @@ -741,11 +741,7 @@ NTSTATUS PhGetMappedImageExportFunction(
         }
         else
         {
    -        Function->Function = PhMappedImageRvaToVa(
    -            Exports->MappedImage,
    -            rva,
    -            NULL
    -            );
    +       Function->Function = (PVOID)((ULONG_PTR) rva);
            Function->ForwardedName = NULL;
         }

dmex:
> phlib map the PE in memory. I don't know why it has an impact on the computed VA, but it does.

The main difference is that the image is backed by real memory (instead of a stack based buffer) and pointers within the image have 'real' addresses, you don't need to readfile, memcpy and parse chunks of the file multiple times to access and display data etc...

> the official PE-COFF spec sheet explicitly say it's a rva

@wj32 would know more about it... I think PhGetMappedImageExportFunction was originally designed with a different purpose hence why it returned different data. 

I've committed your patch for now. It does break plugin compatibility but I don't think anything other than peview was using the function so there shouldn't be any issues changing it 😄

dmex:
> user32.dll is a good example of unnamed exports with invalid VA:
> http://i.imgur.com/3IyrUN0.png

Oops... those entries in that screenshot are in fact exported variables... You can have "unnamed functions" and "unnamed variables" with variables having a VA of 0x0 but are valid at runtime.

PR #133, opened 2017-04-28
phlib: Fix a potential exception when calling NtQueryObject

pbatard:
* Under some circumstances, `NtQueryObject()` will produce a NULL pointer exception if the last parameter is NULL:
image3
* I was able to consistently experience this behaviour, on Windows 10 x64 (Build 15063) when running the `Debug|Win32` version of PH from an __elevated__ VS2017, and then checking the properties of one of the `svchost.exe` services.
* Of course, one is not supposed to run the Win32 version of PH on x64, and as far as I could see, the x64 version does not suffer from this problem, but this is an easy fix to add.

### DIFF
--- a/phlib/hndlinfo.c
+++ b/phlib/hndlinfo.c
@@ -142,12 +142,13 @@ NTSTATUS PhpGetObjectBasicInformation(
     }
     else
     {
+        ULONG returnLength;
         status = NtQueryObject(
             Handle,
             ObjectBasicInformation,
             BasicInformation,
             sizeof(OBJECT_BASIC_INFORMATION),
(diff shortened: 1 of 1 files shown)


dmex:
The documentation for the API (and the sdk headers) specify the ReturnLength parameter as optional and if it's only crashing under wow64 because the parameter is null then it's definitely a bug in the native API that needs to be fixed.

You should probably let Microsoft know about the bug because it's likely crashing other software too.

Thanks for the patch 👍

pbatard:
Thanks! I'll see what I can do to let Microsoft know about that issue.

PR #134, opened 2017-05-03
Add cfg support for wow64 process

lucasg:
Following the remark made by @ionescu007 (https://github.com/processhacker2/processhacker2/pull/124#issuecomment-297221437), I've added the support for tagging memory page which belongs to wow64 CFG bitmaps. Here an example :

sans titre

* Locate and tag `CFG` bitmap memory page (both x86 and x64) for wow64
  processes
* Add `CFG` entries parsing for 32-bit executables in ```peview```
* Add unnamed union in `PH_MAPPED_IMAGE` in order to selectively read a
  `IMAGE_NT_HEADER` as 32 bit (could also have used a C-style cast)
* Rename `CFG` page `"RVA"` column to `"VA"` since we read Virtual Adresses (see PR https://github.com/processhacker2/processhacker2/pull/125)

I've not tested on build `14393`, feel free to tell me if something seem fishy to you.

Cheers,

Lucas.

### DIFF
--- a/ProcessHacker/include/memprv.h
+++ b/ProcessHacker/include/memprv.h
@@ -21,7 +21,8 @@ typedef enum _PH_MEMORY_REGION_TYPE
     Heap32Region,
     HeapSegmentRegion,
     HeapSegment32Region,
-    CfgBitmapRegion
+    CfgBitmapRegion,
+    CfgBitmap32Region
 } PH_MEMORY_REGION_TYPE;
 
 typedef struct _PH_MEMORY_ITEM
--- a/ProcessHacker/memlist.c
+++ b/ProcessHacker/memlist.c
@@ -425,7 +425,9 @@ PPH_STRING PhpGetMemoryRegionUseText(
         return PhFormatString(L"Heap segment%s (ID %u)",
             type == HeapSegment32Region ? L" 32-bit" : L"", (ULONG)MemoryItem->u.HeapSegment.HeapItem->u.Heap.Index + 1);
     case CfgBitmapRegion:
-        return PhFormatString(L"CFG Bitmap");
+    case CfgBitmap32Region:
+        return PhFormatString(L"CFG Bitmap%s",
+            type == CfgBitmap32Region ? L" 32-bit" : L"");
     default:
         return PhReferenceEmptyString();
     }
--- a/ProcessHacker/memprv.c
+++ b/ProcessHacker/memprv.c
@@ -488,23 +488,81 @@ NTSTATUS PhpUpdateMemoryRegionTypes(
     }
 
 #ifdef _WIN64
-    // CFG bitmaps for 64-bit processes
-    if (!isWow64)
+
+    LDR_INIT_BLOCK ldrInitBlock = { 0 };
+    PVOID ldrInitBlockBaseAddress = NULL;
+    PPH_MEMORY_ITEM cfgBitmapMemoryItem;
+    PPH_STRING ntdllFileName;
+    
+    ntdllFileName = PhConcatStrings2(USER_SHARED_DATA->NtSystemRoot, L"\\System32\\ntdll.dll");
+    status = PhGetProcedureAddressRemote(
+        ProcessHandle,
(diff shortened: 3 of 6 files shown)


dmex:
I find this somewhat confusing since the wow64 version of the LDR_INIT_BLOCK structure is the same size as the 64bit version... The structure fields should be ULONG instead of PVOID in order to get the correct addresses from wow64 but it works without those changes!?

@ionescu007 

Any idea why wow64 uses the same structure as 64bit?

dmex:
@ionescu007 

Any idea why the 64bit structure works on build 15063 for wow64 processes but not on build 14393?

ionescu007:
They may have changed the typedef as part of the WowA64 work?

Best regards,
Alex Ionescu

ionescu007:
Or are you reading the 32-bit ntdll's block, instead of the 64-bit's
ntdll's block?

Best regards,
Alex Ionescu

dmex:
It should be reading the block from the 32bit ntdll but it appears to be identical to the 64bit ntdll block (on 15063 anyway)... Not sure if there's a bug or the structure is actually different? 😃

lucasg:
Both block have the same structure in build 15xxx:

image

By the way, here is the 64-bit structure for 15063:

	typedef struct _LDR_INIT_BLOCK{
		ULONG Size;
		ULONG Unkown[2];
		
		struct {
			UINT_PTR LdrInitializeThunk;
			UINT_PTR KiUserExceptionDispatcher;
			UINT_PTR KiUserApcDispatcher;
			UINT_PTR KiUserCallbackDispatcher;
			UINT_PTR RtlUserThreadStart;
			UINT_PTR RtlpQueryProcessDebugInformationRemote;
			UINT_PTR RtlpLegacyApplicationCapabilityNames;
			UINT_PTR LdrSystemDllInitBlock;
		} Wow64Config;
		
		ULONG Unkown2[8];
		ULONG Unkown3; // randomly generated. Looks like a process-wide cookie
		
		DWORD CfgGuardFlags;
		DWORD Unkown4;

		ULONG Unkown5;
		UINT_PTR CfgBitmapAddress;
		ULONG Unkown6; // looks like a size.
		

	} LDR_INIT_BLOCK, *PLDR_INIT_BLOCK;

ionescu007:
After the bitmap there's a SIZE_T which is 2TB And the size of the bitmap.

There should also be a 32MB or 48MB Size, and the Wow64 bitmap address.

Unknown3 is a random number/cookie.

Best regards,
Alex Ionescu

dmex:
I've found the structure in some public symbols and it's called PS_SYSTEM_DLL_INIT_BLOCK.

I reversed the structure based on the 64bit symbols and I've probably used the wrong types.... Can you guys please double check the types for the wow64 version of the structure compared to the below?

```
typedef struct _PS_MITIGATION_OPTIONS_MAP
{
    ULONG_PTR Map[2];
} PS_MITIGATION_OPTIONS_MAP, *PPS_MITIGATION_OPTIONS_MAP;

typedef struct _PS_SYSTEM_DLL_INIT_BLOCK
{
    ULONG Size;
    ULONG_PTR SystemDllWowRelocation;
    ULONG_PTR SystemDllNativeRelocation;
    ULONG_PTR Wow64SharedInformation[16];
    ULONG RngData;
    union
    {
        ULONG Flags;
        struct
        {
            ULONG CfgOverride : 1;
            ULONG Reserved : 31;
        };
    };
    PS_MITIGATION_OPTIONS_MAP MitigationOptionsMap;
    ULONG_PTR CfgBitMap;
    ULONG_PTR CfgBitMapSize;
    ULONG_PTR Wow64CfgBitMap;
    ULONG_PTR Wow64CfgBitMapSize;
} PS_SYSTEM_DLL_INIT_BLOCK, *PPS_SYSTEM_DLL_INIT_BLOCK;
```

ionescu007:
Which PDB had this type?

Best regards,
Alex Ionescu

dmex:
The PDBs are ntkrnlmp.pdb and combase.pdb.

PR #135, opened 2017-05-03
Fix search box font size on high DPI displays

pbatard:
On High DPI displays, the font size used by the _Find Handles or DLLs_ is too big, which makes the ability to enter text difficult:
image3
It therefore seems that the call to `PH_SCALE_DPI()` is unwarranted. This patch removes it.

Tested on Windows 10 x64 Build 15063 (Creators Update) with a default UI scale of 200%.

### DIFF
--- a/ProcessHacker/searchbox.c
+++ b/ProcessHacker/searchbox.c
@@ -83,7 +83,7 @@ VOID PhpSearchInitializeFont(
     if (Context->WindowFont) 
         DeleteObject(Context->WindowFont);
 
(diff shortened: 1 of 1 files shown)


dmex:
My arch-enemy DPI strikes again... Thanks, I'll try sort out the other DPI issues 👍

dmex:
Can you try the latest build and let me know if there are any remaining DPI issues?

pbatard:
I'll try to do that later on today and let you know.

pbatard:
Been playing around with the latest nightly build (while I was experimenting with your process search suggestion in Rufus), and haven't noticed any high DPI issues. Everything looks good at 200% UI scale. 👍

PR #136, opened 2017-05-05
[cfg] Fix typos when parsing CFG infos in PE load config

lucasg:
There was some uninspired copy-pastes when adding support for CFG parsing
for wow64 PE.

### DIFF
--- a/phlib/mapimg.c
+++ b/phlib/mapimg.c
@@ -1277,7 +1277,7 @@ NTSTATUS PhGetMappedImageCfg64(
         CfgConfig->NumberOfGuardAdressIatEntries = config64->GuardAddressTakenIatEntryCount;
         CfgConfig->GuardAdressIatTable = PhMappedImageRvaToVa(
             MappedImage,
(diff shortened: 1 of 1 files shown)


dmex:
I've updated your PR with some additional changes and added settings support to make using peviewer easier.

I found another issue (unrelated to the PR)... If you inspect properties for "C:\Program Files (x86)\Skype\Phone\Skype.exe" you'll see hundreds of delay imports with incorrect values. 

I'm not sure how its parsing imports incorrectly, any ideas?
http://i.imgur.com/ai1ro4j.png

lucasg:
Without having the `PE` to test with, I would say :

* `DelayImport` and `Import` descriptor entries have an `AllAttributes` field member with a flag defining if the import is RVA based or not. This is not taken into account in `mapimg.c`. See radare2 implementation : https://github.com/radare/radare2/blob/c0c0cba3398466661afed716f709f1fc404a752e/libr/bin/format/pe/pe.c#L2141

* There is `BoundImportAddressTableRVA` which is not parsed in `mapimg.c` but it's only an optional flag.

My money on the `AllAttributes`.

PR #139, opened 2017-05-11
Add support for undecorate C++ names

lucasg:
Add support to undecorate C++ names in `phlib`. 
Example : 
    `?_Mtx_unlock@threads@stdext@@YAXPEAX@Z` => `void stdext::threads::_Mtx_unlock(void *)`

Modifs : 

* `phlib` : Add `PhUndecorateName(W)`
* `peviewer` : undecorate names on import and export entries (not on CFG entries since we already rely on the symbol provider)

Peview:

image

NB : there is a MFC bug where the ListView has a hard limit of 260 char to display for text.

### DIFF
--- a/phlib/include/symprv.h
+++ b/phlib/include/symprv.h
@@ -296,6 +296,22 @@ PhWalkThreadStack(
     _In_opt_ PVOID Context
     );
 
+PHLIBAPI
+PPH_STRING
+NTAPI
+PhUndecorateName(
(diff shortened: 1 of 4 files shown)


dmex:
> the ListView has a hard limit of 260 char 

I'll be adding search to peview sometime soon and switch the ListView for our own Treelist control... You'll just have to bear it for now 😄 

> undecorate names on import and export entries

PR merged 👍

PR #189, opened 2017-11-14
Fix some TEB field offsets on 32 bit

Mattiwatti:
After updating to the RS3 types I noticed that the `LastStatusValue` field in the TEB (which is fairly far from the start of the struct) was wrong. After dumping the RS3 structs in WinDbg I arrived at this fix which adds some `#ifdef`s to match the offsets found in the debug symbols.

Going back to earlier kernel PDBs I see that some 'reserved'/padding fields (e.g. `SystemReserved1`) have always had different array sizes on 32 bit compared to 64, so it's possible the offsets were already incorrect prior to this and the RS3 changes are unrelated. I just happened to notice it now.


dmex:
LGTM

Thanks for fixing this 👍

PR #210, opened 2017-12-21
Tagging Apiset memory page

lucasg:
[ApiSet schema map](https://msdn.microsoft.com/en-us/library/windows/desktop/hh802935(v=vs.85).aspx) is located inside every process virtual memory and can be located via the PEB : 

image


dmex:
Thanks 👍

Issue #99, opened 2017-01-16
Private WS are not summed


freMea:
Hi,

When 
`Options>Advanced>Include CPU (and other) usage of children in collapsed processes`
is enabled, while **Working set** values are well summed, this is not the case for **Private WS**.

**Request:**

1. Fix that
2. Add option to use **Private WS** in tray instead of **Working set** for **physical memory**.

Thanks

PR #212, opened 2017-12-27
Add aggregation support to more performance counters

ge0rdi:
I've just discovered this cool PH feature that counters of child processes could be aggregated in parent. Though not all counters are properly aggregated.

So I've added support for aggregation to the rest of counters (where I think it makes sense).

It also fixes issue #99.


dmex:
Looks good. Thanks 😃

PR #215, opened 2017-12-28
Indent background coloring like ProcExp

starkeyvsk:
First github.com pull request... 😬
First time working with Win32/GDI...
Feedback appreciated; I'm sure there is a cleaner/better way to do this, but it's already taken a while to figure out how this works in the first place. 

BTW, It's a neat custom tree control. Perhaps this indentation should be an option controlled in the menu, but I personally prefer it as default. I could look into incorporating it as an option, i haven't had a chance to see how that code works yet though....

### DIFF
--- a/phlib/treenew.c
+++ b/phlib/treenew.c
@@ -5012,6 +5012,14 @@ VOID PhTnpPaint(
 
         PhTnpPrepareRowForDraw(Context, hdc, node);
 
+		int backgroundIndentationAmmount = node->Level * SmallIconWidth;			// amount to indent by level
+		//backgroundIndentationAmmount += (node->s.IsLeaf) * SmallIconWidth;		// if there is a twirldown, indent for that arrow
+		backgroundIndentationAmmount += SmallIconWidth + TNP_CELL_LEFT_MARGIN - 1;  // indent for the icon and margin.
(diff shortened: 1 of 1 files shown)


wj32:
Thanks, but I don't think this belongs in the product.

starkeyvsk:
Looks like so compared to ProcExp:

image

starkeyvsk:
Do you just prefer the full width background?

OK, well I got a question regarding another feature then:
ProcExp will highlight the CPU/PrivateBytes/WorkingSet cells' colors darker if that number goes high, it's a lot easier on the eyes to see when something is stressing the machine. 

Would that be possible to get in  PH3 or are you not looking for any graphical changes to the tree view?

dmex:
@starkeyvsk 

>highlight the colors darker if that number goes high

It's a Heat Map.

> Would that be possible to get in PH3 or are you not looking for any graphical changes to the tree view?

The support is already there for drawing individual cells (TreeNewCustomDraw) as a heat map (just like PH already does for the CPU graphs on the process list). 

A heat map for something like individual cpu cores works very well (something we do need to implement) but doesn't make any sense when the concept is applied to individual processes.

For example benchmarking programs, diagnostic tools, games and various other things will always show a 'dark' color for high system usage but that's exactly what they were designed to do so what's the benefit?

PR #225, opened 2018-01-14
Don't show border in main tabs

ge0rdi:
I'd like to propose this cosmetic change. Border around list/tree views in main tabs looks a bit distracting. The window looks more clean and modern without the border, imho.

Though it is matter of taste, so not sure you will like it.

Original:
ph-main-border
Without border:
ph-main-no-border

### DIFF
--- a/ProcessHacker/mainwnd.c
+++ b/ProcessHacker/mainwnd.c
@@ -394,6 +394,7 @@ VOID PhMwpInitializeControls(
     )
 {
     ULONG thinRows;
+    ULONG treelistBorder;
 
     TabControlHandle = CreateWindow(
         WC_TABCONTROL,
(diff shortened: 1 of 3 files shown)


ProcessHacker/mainwnd.c
```diff
@@ -416,7 +416,7 @@ VOID PhMwpInitializeControls(
     PhMwpProcessTreeNewHandle = CreateWindow(
         PH_TREENEW_CLASSNAME,
         NULL,
-        WS_CHILD | WS_VISIBLE | WS_CLIPCHILDREN | WS_CLIPSIBLINGS | WS_BORDER | TN_STYLE_ICONS | TN_STYLE_DOUBLE_BUFFERED | TN_STYLE_ANIMATE_DIVIDER | thinRows,
```
dmex: You should make this an optional setting. 

1) Open \ProcessHacker\settings.c and add a new setting: PhpAddIntegerSetting(L"TreeListBorderEnable", L"1");

2) Instead of removing WS_BORDER use the following code: 
ULONG treelistBorder;

treelistBorder = PhGetIntegerSetting(L"TreeListBorderEnable") ? WS_BORDER : 0;

mainwnd.c: change "| thinRows" to "| treelistBorder | thinRows"

ge0rdi: Ah, right. Didn't realize that.
Thanks for pointing out.

PR #226, opened 2018-01-15
Get proper information for kernel modules

ge0rdi:
It is not possible to read kernel memory from user-mode. Though we can read headers directly from files (in case of kernel modules).

This will fix "CF Guard", "ASLR" and "Entry point" columns for kernel modules.

### DIFF
--- a/ProcessHacker/modprv.c
+++ b/ProcessHacker/modprv.c
@@ -24,6 +24,7 @@
 #include <modprv.h>
 
 #include <mapimg.h>
+#include <kphuser.h>
 #include <verify.h>
 #include <workqueue.h>
 
@@ -486,7 +487,6 @@ VOID PhModuleProviderUpdate(
             moduleItem->FileName = module->FileName;
 
             PhPrintPointer(moduleItem->BaseAddressString, moduleItem->BaseAddress);
-            PhPrintPointer(moduleItem->EntryPointAddressString, moduleItem->EntryPoint);
 
             PhInitializeImageVersionInfo(&moduleItem->VersionInfo, moduleItem->FileName->Buffer);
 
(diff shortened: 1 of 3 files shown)


dmex:
> It is not possible to read kernel memory from user-mode. 

I can add support for this... What addresses/symbols do you need to read from the kernel for this information?

ge0rdi:
Well, I need to read PE headers from each kernel module to get _ImageBase_, _AddressOfEntryPoint_, _DllCharacteristics_, _TimeDateStamp_ fields.

Though now I see it should be possible to read it using _KphReadVirtualMemoryUnsafe_.
I was looking at its implementation and though it doesn't allow reads from kernel memory. But the check is for output buffer not _base address_. So it should work.

I can try to do it this way.
Basically I could modify _PhLoadRemoteMappedImage_ to use different read function for kernel memory. Or maybe create new _PhLoadRemoteMappedKernelImage_ function?

What do you think?

dmex:
> I could modify PhLoadRemoteMappedImage to use different read function for kernel memory

IMO we should use a new function since it'll prevent any accidental kernel queries from anything currently using PhLoadRemoteMappedImage but if it's 'cleaner' and more reliable to re-use an existing function then go for it.

> it should be possible to read it using KphReadVirtualMemoryUnsafe

If it doesn't work then we can change that behavior to support this feature.

ge0rdi:
Oki. Will do.

> If it doesn't work then we can change that behavior to support this feature.

One question about this. If we'll need to modify _KphReadVirtualMemoryUnsafe_ behavior, we need to change driver. I've noticed that nightly builds still use original driver from 2.39.

Are you still able to _properly_ sign new driver? Because since Win10 RS1 drivers require MS signatures in order to load - well at least on clean installations with _SecureBoot_ enabled (for now).

dmex:
> Are you still able to properly sign new driver?

Yes.

> I've noticed that nightly builds still use original driver from 2.39.

Yep. I'll be making an announcement sometime soon that will shed some light on what's been going on here.

ge0rdi:
Cool :)
Glad to hear that.

ge0rdi:
I've added _PhLoadRemoteMappedImageEx_ that takes pointer to function that is then used to read virtual memory.
For kernel modules I'm calling it with _KphReadVirtualMemoryUnsafe_ as read function.

This basically works well (Win10 RS3 x64).

It doesn't work for paged-out kernel modules. _KpiReadVirtualMemoryUnsafe_ fails when requested region contains non resident pages. I guess this is to avoid unwanted kernel working set pollution..

I think we can improve this in the future. Either modify driver to be able to read paged-out memory. Or we can fall-back to reading headers from files.

dmex:
@ge0rdi 

> It doesn't work for paged-out kernel modules.

I've asked @ionescu007 for some details about this... I'll merge the PR if he agrees with these changes 👍

PR #244, opened 2018-03-02
Patch for user service process (instance)

hsebs:
The registry key of a user service instance doesn't store service dll path. 
It stored in user service template.

The key name of user service instance consists key name of it's template and luid for session.
`[User Service Instance Key Name] = [User Service Template Key Name]_[LUID for Session]`

### DIFF
--- a/ProcessHacker/srvprp.c
+++ b/ProcessHacker/srvprp.c
@@ -279,14 +279,33 @@ INT_PTR CALLBACK PhpServiceGeneralDlgProc(
             SetDlgItemText(hwndDlg, IDC_PASSWORD, L"password");
             Button_SetCheck(GetDlgItem(hwndDlg, IDC_PASSWORDCHECK), BST_UNCHECKED);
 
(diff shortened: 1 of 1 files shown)


hsebs:
This commit redirects an user service instance key name to a user service template key name to get service dll path of a user service instance.

dmex:
@hsebs

How come you closed the PR?

hsebs:
@dmex 
There may be something wrong because I'm not accustomed to git.

I have found more configurations of user service instance that are affected by the user service template. 
I closed it because I thought this commit could cause code fragmentation.

Is it better to reopen the PR?

dmex:
> There may be something wrong because I'm not accustomed to git.

There were no issues with the PR... I just didn't have time to review it until this weekend 😄 

> I closed it because I thought this commit could cause code fragmentation.

You just needed to update the PhGetServiceDllParameter function since that that function is what we use for querying the ServiceDll throughout the program.

I've fixed that up and added you as the co-author 👍 
https://github.com/processhacker/processhacker/commit/a6efcd3ed516945627b01ca5808a2af339beae28

PR #249, opened 2018-03-21
Handling atypical integrity levels

diversenok:
In addition to well-known integrity levels (like High, Medium, and so on) tokens can have intermediate values of mandatory RIDs. For example, Microsoft's `osk.exe` and `Magnify.exe` while running under a limited user (without admin approval mode) like Guest, have integrity levels that are slightly higher than medium. At the moment Process Hacker can't handle such tokens absolutely correctly. For example, you can't change the integrity level of such a process (all menu items are disabled). The idea is to add a new menu item that represents that custom integrity level for such tokens and make other items work correctly.

custom-integrity-level

The reason for the problem with such tokens is that `PhGetTokenIntegrityLevel` can't work with atypical integrity levels (it fails on such tokens) since it simplifies them up to `MANDATORY_LEVEL`. So, I've added a new function `PhGetTokenIntegrityLevelRID` and used it for Integrity menu on the Token tab. In fact, I am not sure if `PhGetTokenIntegrityLevel` is still necessary since `PhGetTokenIntegrityLevelRID` can provide the same information and also can handle the situation described above. Maybe I should've just replaced one function with another.

### DIFF
--- a/ProcessHacker/tokprp.c
+++ b/ProcessHacker/tokprp.c
@@ -922,20 +922,20 @@ INT_PTR CALLBACK PhpTokenPageProc(
                     RECT rect;
                     PPH_EMENU menu;
                     HANDLE tokenHandle;
-                    MANDATORY_LEVEL integrityLevel;
+                    MANDATORY_LEVEL_RID integrityLevelRID;
                     PPH_EMENU_ITEM selectedItem;
 
                     GetWindowRect(GetDlgItem(hwndDlg, IDC_INTEGRITY), &rect);
 
                     menu = PhCreateEMenu();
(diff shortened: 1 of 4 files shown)


ProcessHacker/tokprp.c
```diff
@@ -946,25 +946,38 @@ INT_PTR CALLBACK PhpTokenPageProc(
...
                                     menuItem->Flags |= PH_EMENU_DISABLED;
+                                    customLevelPosition = i + 1;
                                 }
                             }
+
+                            if (customLevelPosition)
+                            {
+                                PPH_EMENU_ITEM unknownIntegrityItem;
+                                
+                                unknownIntegrityItem = PhCreateEMenuItem(0, (ULONG)integrityLevelRID, L"Custom level", NULL, NULL);
```
dmex: Maybe rename the menu "Intermediate level" since the level has an intermediate position?

PR #258, opened 2018-04-15
New properties page for file handles

diversenok:
I've implemented a new page for file handles in their properties. It shows the type of the handle (Pipe / File / Directory / Other), the mode, current position of the file pointer, and the size of the file. And it also allows you to flush the file buffer.

file-handles


dmex:
Looks good. Thanks!

PR #267, opened 2018-05-09
UIAccess field in advanced token properties

diversenok:
I've added a new field to Token's properties tab that displays whether UIAccess flag for the token is enabled. This is a very interesting flag since it allows the process to bypass some UIPI restrictions.


phlib/include/phnativeinl.h
```diff
@@ -1183,6 +1183,40 @@ PhGetTokenIsVirtualizationEnabled(
...
+FORCEINLINE
+NTSTATUS
+PhGetTokenIsUIAccessEnabled(
+    _In_ HANDLE TokenHandle,
+    _Out_ PBOOLEAN IsUIAccessEnabled
+    )
+{
+    NTSTATUS status;
+    ULONG returnLength;
+    DWORD uiAccess;
```
dmex: This needs to be ULONG 😄

phlib/include/phnativeinl.h
```diff
@@ -1183,6 +1183,40 @@ PhGetTokenIsVirtualizationEnabled(
...
+
+    status = NtQueryInformationToken(
+        TokenHandle,
+        TokenUIAccess,
+        &uiAccess,
+        sizeof(DWORD),
```
dmex: Same here

phlib/include/phnativeinl.h
diversenok: Does it? ([1](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379626(v=vs.85).aspx#TokenUIAccess), [2](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/content/ntifs/ne-ntifs-_token_information_class))

Well, there's not much difference, so if you insist... ¯\\\_(ツ)\_/¯

dmex: NT uses ULONG for unsigned long types
Win32 uses DWORD for unsigned long types

These are NT kernel types and the wdk headers for those structures are using ULONG... Whomever wrote those MSDN documents is someone from the Win32 team using the incorrect terminology.

diversenok: Ok, I'll take that into account. :+1:

dmex:
Thanks! I usually fixup type/naming stuff without saying anything but I'm working on some other stuff :)

PR #294, opened 2018-07-29
Add .editorconfig file

ge0rdi:
VS2017 finally supports [.editorconfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options) file that can be used to enforce coding consistency.

It is especially useful for people that contribute to more projects, while each project uses different indent style (tab vs spaces, tab size).
`.editorconfig` file overrides VS editor settings, so that there is no need to adjust them when switching between projects.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root=true
+
+[*]
+indent_style=space
+indent_size=4
+trim_trailing_whitespace=true
+insert_final_newline=true


dmex:
> VS2017 finally supports 

Thanks 😃

gvanem:
And what does this mean for us who have an editor that does *not* support an `.editorconfig`?

I guess the chances are that the sources could look like crap. Like indenting with mixed tabs and spaces. This is especially annoying in an editor with visible white-space. Like working with GNU-makefiles, where rules **must** start with a `<TAB>`.  

Or in many GNU project, where Emacs/Vim rules. Like this example from GNU-diff in my Crimson editor:
gnu-diff-mixed-indent

Nice? I don't think so. How this project will keep a consistent coding-style no matter what any `.editorconfig` says. Maybe a [script](https://github.com/curl/curl/blob/master/lib/checksrc.pl) could check for these things before a merge?

ge0rdi:
> And what does this mean for us who have an editor that does not support an .editorconfig?

There will be no change for you.
You will have to make sure to use proper indenting when submitting PR (it was always the case).

> I guess the chances are that the sources could look like crap. Like indenting with mixed tabs and spaces. 

This is up to maintainers (reviewers) to check whether PR contains proper formatting.
Right now it seems to be pretty consistent (4 spaces).

> I hope this project will keep a consistent coding-style no matter what any .editorconfig says.

I definitely agree.
Consistent usage of tabs/spaces is much more important that the eternal question whether to use tabs or spaces.

`.editorconfig` just makes it easier for people using editor compatible with it.

PR #301, opened 2018-08-15
Cast const empty string to PWSTR

yupi2:
`Error C2440 'return': cannot convert from 'const wchar_t [1]' to 'PWSTR'`

I don't think the error/warning is reproducible since it probably would've been noticed already...


dmex:
Why does this line generate a warning out of dozens of functions with the same return value? 🤔

PR #310, opened 2018-09-12
Improving the appearance of groups for tokens.

diversenok:
If a token contains groups that are enabled but not marked with "enabled by default" the result looks a bit strange as if something is broken. I've added a background color for them by analogy with privileges and "Enabled" to the status column. And since groups might have several attributes simultaneously the description column now shows them all.

group-coloring

### DIFF
--- a/ProcessHacker/tokprp.c
+++ b/ProcessHacker/tokprp.c
@@ -85,6 +85,16 @@ typedef struct _TOKEN_PAGE_CONTEXT
     ATTRIBUTE_TREE_CONTEXT AuthzTreeContext;
 } TOKEN_PAGE_CONTEXT, *PTOKEN_PAGE_CONTEXT;
 
+PH_ACCESS_ENTRY GroupDescriptionEntries[6] =
+{
+{ NULL, SE_GROUP_INTEGRITY | SE_GROUP_INTEGRITY_ENABLED, FALSE, FALSE, L"Integrity" },
+{ NULL, SE_GROUP_LOGON_ID, FALSE, FALSE, L"Logon Id" },
+{ NULL, SE_GROUP_OWNER, FALSE, FALSE, L"Owner" },
+{ NULL, SE_GROUP_MANDATORY, FALSE, FALSE, L"Mandatory" },
+{ NULL, SE_GROUP_USE_FOR_DENY_ONLY, FALSE, FALSE, L"Use for deny only" },
+{ NULL, SE_GROUP_RESOURCE, FALSE, FALSE, L"Resource" }
+};
+
 INT CALLBACK PhpTokenPropPageProc(
     _In_ HWND hwnd,
     _In_ UINT uMsg,
@@ -246,7 +256,7 @@ PPH_STRING PhGetGroupAttributesString(
         if (Attributes & SE_GROUP_ENABLED_BY_DEFAULT)
             string = PhCreateString(L"Default Enabled");
         else if (Attributes & SE_GROUP_ENABLED)
-            string = PhReferenceEmptyString();
+            string = PhCreateString(L"Enabled");
         else
             string = PhCreateString(L"Disabled");
     }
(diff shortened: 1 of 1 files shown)


dmex:
@diversenok 

Do you think reversing the 'enabled' colors would be more useful?

Enabled: Dark Green
Default: Light Green

diversenok:
Hmm... I don't think so. "Default Enabled" is more mandatory than plain "Enabled" since it can't be disabled using `AdjustTokenGroups` or `AdjustTokenPrivileges`. And, I guess, deeper colors look more like "mandatory" ones. =)

dmex:
>  "Default Enabled" is more mandatory than plain "Enabled" since it can't be disabled 

That's true for the current highlighting regime but I was thinking of changing this since the defaults are always going to be well-known ahead of time but they're highlighted better than the the entries added/removed/modified by the process.

I was thinking the darker colors would be better for non-defaults since they stand out better?

dmex:
> PH should especially highlight modified privileges/groups as we now know the initial state of each entry.

Yep. Highlighting modified entries makes it easier to review/audit tokens 👍

> just clear "Enabled" bit without clearing "Enabled by default" bit

Good catch!

>  "Enabled by default" flag is very straightforward — the only thing it means is that this particular privilege was enabled when the token was created.

This makes a lot more sense. Typically if you wanted to identify entries that are not the defaults you would have to create a  token for that user and compare it against the token for the process BUT you can't create tokens used by system processes without SeCreateTokenPrivilege and/or using a LocalSystem service and so the "Enabled by default" flag is used for that exact purpose.

> But I don't know how to implement this in UI. Maybe we should add a new column that represents default values.

A new column would be the easiest method for now (so we can always see the defaults - even when highlighting is disabled) and we can always change it later.

The only issue that I've found with PR #311 is when disabling privileges that have already been modified.

1) Start the defrag service.
2) View the token information:
https://i.imgur.com/mwr7kCY.png

If you disable entries that have already been modified they're no longer considered modified:
https://i.imgur.com/PdjoUDp.png

It would be nice to have 'modified' when some 3rd party program like PH disables the privilege but if it can't be done then don't worry about it 👍

diversenok:
> You can't create tokens used by system processes without SeCreateTokenPrivilege and/or using a LocalSystem service

SeCreateTokenPrivilege is a bit an overkill — it's necessary only if you want to create a fully customized token from the scratch (and this privilege is also too rare). SeTcbPrivilege should be quite enough since it allows querying any session's token using `WTSQueryUserToken` and obtaining LocalService, NetworkService, and System tokens by logging onto them with `LogonUser`.

> It would be nice to have 'modified' when some 3rd party program like PH disables the privilege

I don't think it's possible without the use of system auditing. I mean, of course, we can query token's information every now and then just trying to catch it in a non-default state (and we can even optimize it using token's Modified ID), but it seems to be a bit noneffective. So, I guess, at this point, we should dwell on the idea that "modified" just means "not default".

By the way, I also have several suggestions and ideas about tokens:

 - What about "Reset" context menu item for privileges that returns them to their initial state?

 - It might be useful to have a visual tracking of tokens with potentially dangerous flags like UIAccess and SandboxInert. The first one allows the process to mess up with UIPI (install global windows hooks that ignore integrity levels, and so on), the second one allows it to bypass Software Restriction Policies and AppLocker. What do you think about adding new entries to the ListView (into a separate group) that only appear when these flags present in the token (to draw attention to their presence)?

 - I also can implement group adjustment. It is actually great to be capable of disabling Administrator or `NT SERVICE\something` membership for a process running under System account. But on the default configuration of Windows, it's almost everything you can do with this feature. You need to have some programs that use custom tokens to take full advantage of it. That's why I am not sure.

dmex:
> What about "Reset" context menu item for privileges that returns them to their initial state?

Yeah this would be great. If you submit the PR then I'll merge it 👍 

> What do you think about adding new entries to the ListView (into a separate group) that only appear when these flags present in the token (to draw attention to their presence)?

I dunno about seperate groups for this case but we should definitely highlight the token and or the process if these flags are enabled.

> I also can implement group adjustment.

This would also be very handy (even if it doesn't always work) especially for developers testing their programs during development and for security auditing. Submit a PR and I'll merge it 😃

PR #311, opened 2018-09-12
Highlight modified privileges and groups

diversenok:
Continuation of #310 (apparently I can't add commits to a pull request that is already merged).

- Reverse background colors
- Make possible to disable privileges that were enabled by default.

### DIFF
--- a/ProcessHacker/tokprp.c
+++ b/ProcessHacker/tokprp.c
@@ -253,12 +253,20 @@ PPH_STRING PhGetGroupAttributesString(
     }
     else
     {
-        if (Attributes & SE_GROUP_ENABLED_BY_DEFAULT)
-            string = PhCreateString(L"Default Enabled");
-        else if (Attributes & SE_GROUP_ENABLED)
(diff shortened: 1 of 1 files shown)


VictorVG:
Ok, but is truncated some string's - **KLASS** in to forum.ru-board.com wrote:

> ...there are two lines of code:
> Disabled (modified
> Disabled (modified)
> those. in the first case, the closing bracket is omitted, so this is the line with no parentheses and is displayed in properties
> svchost (not all, for example, DNS client, DHCP client, etc.) on the Token tab in the Status column

yes, a check and confirmed this confusion. Please, see  scrinshot's:

ph-301862_win10

Win10 LTSB

ph-301862-win7sp1

Win 7 SP1

dmex:
@VictorVG 

Closed pull requests are not the place for reporting bugs... Use the correct place or we won't fix it.

VictorVG:
Ok! I just did not want to start issues - I thought that we would just correct this trifle and forget about it.

PR #312, opened 2018-09-12
Support listing process DPI awareness on pre-8.1 Windows versions.

### LINKED ISSUE #309: Can we have DPI Awareness support pre Windows 8.1?
GetProcessDpiAwareness is only available from Windows 8.1. Before that you had to use IsProcessDPIAware, but that only applies to the current process.

Creating remote threads to call it in other processes is probably too expensive to be a good idea when it is shown in the process list view. The pre-Windows 8.1 version works by simply reading the global char gfDPIAware in user32.dll. The function just consists of a simple movzx, ret, so maybe we could just read the offset out of the instruction when it matches the expected pattern and then NtReadVirtualMemory it when needed.

Thoughts?

### LINKED ISSUE #313: DPI Awareness does not update
Changes to DPI awareness are not reflected by the DPI awareness column, you have to restart PH to see it.

poizan42:
Here is my proposal for fixing #309. 

~~It currently mostly works, but there is an issue with getting information from processes started after process hacker - I think it might be because the information is being read too soon.~~

Fixes #309. Fixes #313.

### DIFF
--- a/ProcessHacker/proctree.c
+++ b/ProcessHacker/proctree.c
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
     }
 }
 
+static ULONG PhpGet_gfDPIAwareOffset32(
+    /* user32VBase is the base the offset in the movzx is relative to,
+     * either the current load address if relocations are performed, or
+     * the default base if not. */
+    _In_ ULONG user32VBase,
+    _In_ PBYTE isProcessDPIAware32)
+{
+    // 0FB605xxxxxxxx    movzx eax, gfDPIAware
+    // C3                ret
+    __try
+    {
+        if (isProcessDPIAware32[0] == 0x0F && isProcessDPIAware32[1] == 0xB6 &&
+            isProcessDPIAware32[2] == 0x05 && isProcessDPIAware32[7] == 0xC3)
+        {
+            return *(PULONG)(isProcessDPIAware32 + 3) - user32VBase;
+        }
+    }
+    __except (EXCEPTION_EXECUTE_HANDLER)
+    {
+        return 0;
+    }
+    return 0;
+}
+
+static ULONG_PTR PhpGet_gfDPIAwareOffset64(
+    _In_ PBYTE user32Base,
+    _In_ PBYTE isProcessDPIAware64)
+{
+    // 0FB605xxxxxxxx    movzx eax,byte [rel gfDPIAware]
+    // C3                ret
+    __try
+    {
+        if (isProcessDPIAware64[0] == 0x0F && isProcessDPIAware64[1] == 0xB6 &&
+            isProcessDPIAware64[2] == 0x05 && isProcessDPIAware64[7] == 0xC3)
+        {
+            /* NB: x86_64 relative addressing is relative to start
+                   of the following instruction, hence the + 7 */
+            PBYTE gfDPIAwareOffsetAddr =
+                PTR_ADD_OFFSET(
+                    PTR_ADD_OFFSET(isProcessDPIAware64, *(PULONG)(isProcessDPIAware64 + 3)),
+                    7);
+            return gfDPIAwareOffsetAddr - user32Base;
+        }
+    }
+    __except (EXCEPTION_EXECUTE_HANDLER)
+    {
+        return 0;
+    }
+    return 0;
+}
+
+static void PhpInit_gfDPIAwareOffset32WoW(_Out_ PULONG_PTR gfDPIAwareOffset32)
+{
+    PBYTE isProcessDPIAware32;
+    PH_STRINGREF systemRoot;
+    PPH_STRING user32FileName;
+    PH_MAPPED_IMAGE mappedImage;
+    PH_MAPPED_IMAGE_EXPORTS exports;
+    PH_MAPPED_IMAGE_EXPORT_FUNCTION isProcessDPIAwareFun;
+    NTSTATUS status;
+
+    *gfDPIAwareOffset32 = 0;
+
+    PhGetSystemRoot(&systemRoot);
+    user32FileName = PhConcatStringRefZ(&systemRoot, L"\\SysWow64\\user32.dll");
+    status = PhLoadMappedImage(user32FileName->Buffer, NULL, TRUE, &mappedImage);
+    PhDereferenceObject(user32FileName);
+    if (!NT_SUCCESS(status))
+        return;
+
+    if (!NT_SUCCESS(PhGetMappedImageExports(&exports, &mappedImage)))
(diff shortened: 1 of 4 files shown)


poizan42:
So fixing #313 also fixes this.

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
     }
 }
 
+static ULONG PhpGet_gfDPIAwareOffset32(
```
dmex: Naming scheme doesn't follow project conventions

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
...
+     * either the current load address if relocations are performed, or
+     * the default base if not. */
+    _In_ ULONG user32VBase,
+    _In_ PBYTE isProcessDPIAware32)
+{
+    // 0FB605xxxxxxxx    movzx eax, gfDPIAware
+    // C3                ret
+    __try
+    {
+        if (isProcessDPIAware32[0] == 0x0F && isProcessDPIAware32[1] == 0xB6 &&
```
dmex: This is a maintenance nightmare.

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
...
+        }
+    }
+    __except (EXCEPTION_EXECUTE_HANDLER)
+    {
+        return 0;
+    }
+    return 0;
+}
+
+static ULONG_PTR PhpGet_gfDPIAwareOffset64(
```
dmex: Same as above

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
...
+        }
+    }
+    __except (EXCEPTION_EXECUTE_HANDLER)
+    {
+        return 0;
+    }
+    return 0;
+}
+
+static void PhpInit_gfDPIAwareOffset32WoW(_Out_ PULONG_PTR gfDPIAwareOffset32)
```
dmex: Naming scheme doesn't follow project conventions

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
...
+        0,
+        &isProcessDPIAwareFun
+        )) || !isProcessDPIAwareFun.Function)
+        goto CleanupExit;
+    isProcessDPIAware32 = PhMappedImageRvaToVa(
+        &mappedImage,
+        PtrToUlong(isProcessDPIAwareFun.Function),
+        NULL);
+    if (isProcessDPIAware32)
+        *gfDPIAwareOffset32 = PhpGet_gfDPIAwareOffset32(
```
dmex: Naming scheme doesn't follow project conventions

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
...
+#define PhpGet_gfDPIAwareOffset PhpGet_gfDPIAwareOffset32
+#endif
+
+static void PhpInitDpiAwarenessValues(
+    _Out_ PVOID* getProcessDpiAwarenessInternal,
+#ifdef _WIN64
+    _Out_ PULONG_PTR gfDPIAwareOffset32,
+#endif
+    _Out_ PULONG_PTR gfDPIAwareOffset
+    )
```
dmex: Naming scheme doesn't follow project conventions

ProcessHacker/proctree.c
```diff
@@ -1211,6 +1211,133 @@ static VOID PhpUpdateProcessNodeAppId(
...
+    if (*getProcessDpiAwarenessInternal =
+        PhGetDllBaseProcedureAddress(user32Base, "GetProcessDpiAwarenessInternal", 0))
+        return;
+    isProcessDPIAware = PhGetDllProcedureAddress(L"user32.dll", "IsProcessDPIAware", 0);
+
+    if (isProcessDPIAware)
+        *gfDPIAwareOffset = PhpGet_gfDPIAwareOffset(user32Base, isProcessDPIAware);
+
+#ifdef _WIN64
+    PhpInit_gfDPIAwareOffset32WoW(gfDPIAwareOffset32);
```
dmex: Naming scheme doesn't follow project conventions

ProcessHacker/proctree.c
```diff
@@ -1220,24 +1347,86 @@ static VOID PhpUpdateProcessNodeDpiAwareness(
         _In_ HANDLE hprocess,
         _Out_ ULONG *value
         );
+    static ULONG_PTR gfDPIAwareOffset;
+    BOOL query_gfDPIAware32 = FALSE;
+    BOOL query_gfDPIAware64 = FALSE;
```
dmex: Naming scheme doesn't follow project conventions

ProcessHacker/proctree.c
poizan42: What would you suggest? PhpGetGfDPIAwareOffset32? The variable is called gfDPIAware so I didn't want to uppercase the g.

ProcessHacker/proctree.c
poizan42: Could you clarify?

ProcessHacker/proctree.c
```diff
@@ -1380,57 +1380,54 @@ static VOID PhpUpdateProcessNodeDpiAwareness(
             return;
     }
 
-    if (!(ProcessNode->ValidMask & PHPN_DPIAWARENESS))
```
dmex: This doesn't actually remove the flags.

dmex:
Please see the project conventions here:
https://github.com/processhacker/processhacker/blob/master/HACKING.md

Even if we ignore the project convention issues the disassembly is ugly and won't be allowed.

You should implement this as a plugin-extra here (you can also ignore the convention issues there):
https://github.com/processhacker/plugins-extra

ProcessHacker/proctree.c
poizan42: I'm not sure if I follow you. Do you want PHPN_DPIAWARENESS to be removed completely? I wasn't sure if it was being used somewhere else.

ProcessHacker/proctree.c
dmex: We don't use these type of workarounds anywhere else in the codebase because that code will break every other time Microsoft patches something and since most other contributors would not know how to fix that code it would just be some poor soul (me) having to fix it for months/years into the future.

PR #326, opened 2018-10-02
Adjustment of groups for tokens

diversenok:
I've implemented resetting of privileges and adjusting of groups in tokens. Besides this, I've moved all restricting SIDs to a separate category since they can't be modified and are treated by the OS as the second group list that can contain arbitrary entries (not necessarily corresponding to token groups).

### DIFF
--- a/ProcessHacker/resource.h
+++ b/ProcessHacker/resource.h
@@ -741,6 +741,10 @@
 #define ID_MINIINFO_REFRESHAUTOMATICALLY 40289
 #define IDC_MAXSCREEN                   40293
 #define ID_EMPTY_COMBINEMEMORYLISTS     40295
+#define ID_PRIVILEGE_RESET              40296
+#define ID_GROUP_ENABLE                 40297
+#define ID_GROUP_DISABLE                40298
+#define ID_GROUP_RESET                  40299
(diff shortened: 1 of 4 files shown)


dmex:
@diversenok 

You can enable 'mandatory labels' which seems a bit weird?

diversenok:
@dmex 

Well, a bit. Mandatory labels are still ordinary SIDs. They are always enabled as integrity level sources, but since they don't have 'use for deny only' flag they can be enabled as groups too. For example, you can add an ACL entry that grants full access to 'medium mandatory level':

    icacls test.txt /grant *S-1-16-8192:F

If you don't have any other permissions on this file it opens successfully only when this mandatory label is enabled _as a group_ in the token of the caller. A bit confusing, but logical.

dmex:
@diversenok 

Are these changes supposed to highlight the Token > Advanced > Capabilities entries?
https://i.imgur.com/Kq37lxo.png

diversenok:
I don't know anything about token capabilities yet. I am researching Windows API, and I'm still at the stage of Windows 7 ;)

If you are talking about that they are considered as modified while they are not... Maybe that's a side effect of the new highlighting, I haven't tested it on Windows 10.

PR #328, opened 2018-10-05
Improve search of memory regions

diversenok:
I've slightly changed the search algorithm used in the Memory page. Now it can handle arbitrary pointers and show you the region to which this pointer belongs.


dmex:
> show you the region to which this pointer belongs

This is a good idea 👍

PR #343, opened 2018-11-29
Retpoline support

TheEragon:
New Windows (spring 2019) will have retpoline support. This patch will display whether the data are present in the PE or not.


dmex:
👍

PR #353, opened 2018-12-20
peview: Add PE Image Machine Type support ARM64 and Hybrid PE

fcharlie:
Add ARM64 and Hybrid PE support.


fcharlie:
@dmex Great.

PR #374, opened 2019-01-30
Add support for environment variables starting with equality sign

diversenok:
In spite of the fact that WinAPI documentation states that the name of an environment variable can't contain an equality sign, it's not entirely true. The name can start with it. Actually, both `GetEnvironmentVariable` and `SetEnvironmentVariable` work fine with such variables.  What about showing all of them in a separate category?

substitutions

I wasn't able to find out whether such class of variables has a particular name. So, I came up with calling them substitutions because of their only known application. These variables (the ones containing a drive letter) are used as current directories on a per-drive basis. They are mostly utilized by the Command Processor, but, actually, WinAPI takes them into account too. The following code will open `D:\Folder\file.txt` since `CreateFile` looks for an environment variable used as a label.

    SetEnvironmentVariable("=D:", "D:\Folder");
    CreateFile("D:file.txt", …)

Thus, sometimes it might be useful to view or change such variables for other processes.

### DIFF
--- a/ProcessHacker/include/procprpp.h
+++ b/ProcessHacker/include/procprpp.h
@@ -347,7 +347,9 @@ typedef struct _PH_ENVIRONMENT_CONTEXT
             ULONG HighlightProcessEnvironment : 1;
             ULONG HighlightUserEnvironment : 1;
             ULONG HighlightSystemEnvironment : 1;
-            ULONG Spare : 25;
+            ULONG HideSubstitutionEnvironment : 1;
+            ULONG HighlightSubstitutionEnvironment : 1;
+            ULONG Spare : 23;
         };
     };
 
--- a/ProcessHacker/prpgenv.c
+++ b/ProcessHacker/prpgenv.c
@@ -35,9 +35,11 @@
 
 typedef enum _ENVIRONMENT_TREE_MENU_ITEM
 {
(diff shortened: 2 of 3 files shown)


dmex:
> What about showing all of them in a separate category?

You could set these per-process/user and machine so you would still need to know where those values come from.

> I wasn't able to find out whether such class of variables has a particular name.

These are internal variables used by cmd.exe for scripts and various CRT functions. One such use of these variables are the per-drive current directory from DOS 1.0:
https://blogs.msdn.microsoft.com/oldnewthing/20100506-00/?p=14133
https://blogs.msdn.microsoft.com/oldnewthing/20101011-00/?p=12563
https://unix.stackexchange.com/a/251215

dmex:
I've modified the PR and included it with commits 19e235d and 38d98a2

I also commented out these lines for now since it might be handy to know which processes are initializing these variables after using functions or launching cmd?
https://github.com/processhacker/processhacker/blob/master/ProcessHacker/prpgenv.c#L242

diversenok:
You are right, that's a better approach.

> I also commented out these lines for now since it might be handy to know which processes are initializing these variables after using functions or launching cmd?

It looks like Explorer is responsible for this one. Some shell API function passes a string like `::{an-arbitrary-guid}` to a file-related API, and it creates a default per-drive current directory on what it considers as a drive.

https://googleprojectzero.blogspot.com/2016/02/the-definitive-guide-on-win32-to-nt.html

PR #402, opened 2019-04-03
Don't prepend "Computer" when open key

xiaoyinl:
When user clicks "Open key", the key name is prepended "Computer", but Regedit doesn't recognize "Computer" if the system locale is not English.

I found that at least on Windows 10 17763, there's no need to prepend "Computer\\". Regedit can open the correct key when the LastKey or Favorites values start with "HKEY_LOCAL_MACHINE", "HKEY_CURRENT_USER", etc.


dmex:
> Regedit doesn't recognize "Computer" if the system locale is not English.

Good catch, thanks! 😃

PR #404, opened 2019-04-03
Mark terminated thread handles

diversenok:
Sometimes it is useful to determine who's leaking handles to terminated threads. I suggest labeling such handles in the same way as is already done for non-existent processes.

I also added exit status field into thread handle properties.

Remarks:
The best method for determining the existence of a thread is to search for it in the last process-thread snapshot. It fits perfectly to the implementation of `PhStdGetClientIdName`. However, I am not sure how to do it inside `PhGetClientIdNameEx`, so I've chosen the easy way which might be not as scalable as desired. :thinking:


dmex:
We need to find a method without PhStdGetClientIdName. I'm going to close for now because of the conflicts.

PR #405, opened 2019-04-05
Fix passive wait analysis on Windows 7/8

Mattiwatti:
`NtQueryInformationThread(ThreadLastSystemCall)` currently always fails on Windows 7/8 with `STATUS_INFO_LENGTH_MISMATCH` due to the `WaitTime` member, which was only added in 8.1. This makes passive thread wait analysis fail in PH.

I've moved the `WaitTime` field inside an `#if` version check, but note that the call will still succeed on Windows 8.1 and 10 without this field, since `PspQueryLastCallThread` accepts both versions of the struct.

Re: the `Pad` fields: I've removed these since the sizes and offsets are already correct due to compiler inserted padding for both x86 and x64.

### DIFF
--- a/phnt/include/ntpsapi.h
+++ b/phnt/include/ntpsapi.h
@@ -911,12 +911,9 @@ typedef struct _THREAD_LAST_SYSCALL_INFORMATION
 {
     PVOID FirstArgument;
     USHORT SystemCallNumber;
-#ifdef WIN64
-    USHORT Pad[0x3]; // since REDSTONE2
-#else
-    USHORT Pad[0x1]; // since REDSTONE2
+#if (PHNT_VERSION >= PHNT_WINBLUE)
+    ULONG64 WaitTime; // may be omitted
 #endif
-    ULONG64 WaitTime;
 } THREAD_LAST_SYSCALL_INFORMATION, *PTHREAD_LAST_SYSCALL_INFORMATION;
 
 // private


dmex:
> NtQueryInformationThread(ThreadLastSystemCall) currently always fails on Windows 7/8 with STATUS_INFO_LENGTH_MISMATCH

PhGetThreadLastSystemCall passes the length required for Windows 7.

For example:
https://github.com/processhacker/processhacker/blob/a39d0f312c863ccb1c4ed927bfcd8eedac1b5fd7/phlib/include/phnativeinl.h#L889-L895


> This makes passive thread wait analysis fail in PH.

Works for me? 
https://i.imgur.com/wQqiv8h.png

> Re: the Pad fields: I've removed these since the sizes and offsets are already correct due to compiler inserted padding for both x86 and x64.

The Pad fields were needed with RTL_SIZEOF_THROUGH_FIELD for passing the correct length on Windows 7 while keeping the structure updated for Windows 10.

> But combase.pdb has that Pad[3] member in that struct.

64bit symbols don't generally include padding but this one does for some reason.

Mattiwatti:
My bad. The reason the wait analysis didn't work for me is because I was using a repository that for some reason didn't have 3ffb71e (I cherry pick a lot of commits). This commit fixed the issue by calling `PhGetThreadLastSystemCall` instead of `NtQueryInformationThread`.

I'm not sure I agree with using `RTL_SIZEOF_THROUGH_FIELD` basically ever, since it will always miss alignment padding and this way the padding has to muddy up the headers, but that wasn't the point of the PR anyway. Will close this.

dmex:
> I was using a repository that for some reason didn't have 3ffb71e (I cherry pick a lot of commits).

Fair enough. 

> I'm not sure I agree with using RTL_SIZEOF_THROUGH_FIELD basically ever

There are some cases where it makes sense but not very many. I use it because it's what Windows uses for accessing undeclared/undefined fields at the end of structures and for getting backwards compatible sizes.

For instance you won't find code for accessing the DirectoryEntries field of the IMAGE_RESOURCE_DIRECTORY structure on Windows doing this:
`(char*)resourceDirectory + 1`

Instead it uses the macro with the base address:
`RTL_SIZEOF_THROUGH_FIELD(IMAGE_RESOURCE_DIRECTORY, NumberOfIdEntries)`

> this way the padding has to muddy up the headers

The THREAD_LAST_SYSCALL_INFORMATION structure is the one exception where I've included the Padding but only because the padding was defined in 64bit symbols.

PR #437, opened 2019-07-10
Fix some potential null dereferences

diversenok:
PH used to crash while viewing detailed information about a handle of a terminated process or a thread that exited with a code with no textual description.

Looks like PhGetStatusMessage, PhGetWin32Message, and PhGetNtMessage can return NULL which gets  dereferenced in several places in the code.

### DIFF
--- a/ProcessHacker/hndlprp.c
+++ b/ProcessHacker/hndlprp.c
@@ -1064,9 +1064,9 @@ VOID PhpUpdateHandleGeneral(
 
                 status = PhGetStatusMessage(exitStatus, 0);
                 exitcode = PhFormatString(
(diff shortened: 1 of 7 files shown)


dmex:
Good catch but the PR added more issues. 

For example:
PhGetString((PPH_STRING)PH_AUTO(PhGetWin32Message(win32Result))))->Buffer);

PhGetMessage returns NULL but PH_AUTO attempts to free and crashes or PhGetString returns NULL which then gets used in printf.

dmex:
Github also didn't add you as a contributor to the commit even though you showed up in the desktop version 😕

PR #440, opened 2019-07-14
make RunFileDialog showing 'Run as Administrator' configurable

RoccoMatano:
Since I have been using the RunFileDialog of ProcessHacker 2.39 a lot in the past years, I've gotten used to the fact that it ran programs as administrator.

Now I tried to update to the latest nightly build of ProcessHacker 3.0 and had to realize that now there is this extra checkbox that has to be clicked in order to run programs as administrator.

That is why I like to propose to introduce a setting that lets me decide whether to have this extra checkbox or not. In this pull request the default to display it. And there is no UI to change that. But I added a setting so that the checkbox can be disable manually.

### DIFF
--- a/ProcessHacker/mainwnd.c
+++ b/ProcessHacker/mainwnd.c
@@ -605,8 +605,13 @@ VOID PhMwpOnCommand(
         break;
     case ID_HACKER_RUN:
         {
+            ULONG flags = RFF_OPTRUNAS;
+            if (PhGetIntegerSetting(L"RunDlgHideRunAsAdministrator"))
+            {
+                flags = 0;
+            }
             SelectedRunAsMode = 0;
-            PhShowRunFileDialog(WindowHandle, NULL, NULL, NULL, NULL, RFF_OPTRUNAS);
+            PhShowRunFileDialog(WindowHandle, NULL, NULL, NULL, NULL, flags);
         }
         break;
     case ID_HACKER_RUNASADMINISTRATOR:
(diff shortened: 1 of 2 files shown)


dmex:
@RoccoMatano 

I've added a new custom built RunFileDialog in the latest nightly and the 'Run as administrator' checkbox will remember its previous state when you open/close the dialog.

Since we're now using our own dialog instead of the shell32 dialog we can change the layout/make things more customizable... do you think it should still be possible to hide the checkbox or are there other changes that we could add?

RoccoMatano:
@dmex 

My intention was not so much to hide the checkbox, but to free myself from having to set the check mark each and every time that I used the RunFileDialog to start an elevated process. Since now there is the setting "RunFileDlgState" which controls exactly whether the user has to do this or not, I think the setting for hiding the checkbox has become uncalled-for.

(Currently I am just looking at the code. In case I should have overlooked anything, I will provide a PR).

RoccoMatano:
@dmex 

Now I had the opportunity to actually run the latest build. While the checkbox works as expected, the main RunFileDialog functionality has a problem:
When entering the name of a 'real' executable (i.e. *.exe) or batch file (i.e. *.bat or *.cmd) it works as expected, but although I extended my %PATHEXT% variable to include python files (*.py) and associated those with python.exe, those are not accepted (i.e. nothing happens, when hitting 'OK').
It is quite clear where this problem arises: PhParseCommandLineFuzzy only ever searches the %PATH% for filenames that are missing the '.exe' extension. That's not enough - all entries of %PATHEXT% should be tried until a match is found.

RoccoMatano:
@dmex

I don't quite get why it is tried to get the full file path in the case, where ShellExecuteEx is used. As long as the command line is split into file name and parameters ShellExecuteEx will do the 'normal' file search anyway.

dmex:
@RoccoMatano 

> the main RunFileDialog functionality has a problem

There are a few problems but its expected. The built-in RunFileDlg has a 30 year head start and there are probably a few things missing.

> ShellExecuteEx will do the 'normal' file search anyway.

I just re-used some existing code to make things work with the initial version. I'll keep improving it over time but feel free to submit any patches if you can't wait 👍

PR #487, opened 2019-10-21
phnt improvements

diversenok:
Here are some small improvements for phnt headers.


phnt/include/ntioapi.h
```diff
@@ -417,7 +417,14 @@ typedef struct _FILE_MOVE_CLUSTER_INFORMATION
 
 typedef struct _FILE_RENAME_INFORMATION
 {
+#if (PHNT_VERSION >= PHNT_REDSTONE)
+    union {
+        BOOLEAN ReplaceIfExists;  // FileRenameInformation
```
dmex: RE above comment

diversenok: I just copied it from ntifs.h, but I agree, -EX version suits better.

PR #488, opened 2019-10-21
Detect terminated threads correctly

diversenok:
If a thread exits with STATUS_PENDING, Process Hacker assumes it's still alive, while it is not. Fix this behavior by using the `ThreadIsTerminated` information class.

### DIFF
--- a/ProcessHacker/hndlprp.c
+++ b/ProcessHacker/hndlprp.c
@@ -1100,7 +1100,7 @@ VOID PhpUpdateHandleGeneral(
 
         if (NT_SUCCESS(status))
         {
-            ULONG isTerminated = FALSE;
+            BOOL isTerminated = FALSE;
             THREAD_BASIC_INFORMATION basicInfo;
             KERNEL_USER_TIMES times;
             PPH_STRING name;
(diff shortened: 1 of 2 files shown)


ProcessHacker/hndlprp.c
```diff
@@ -1100,7 +1100,7 @@ VOID PhpUpdateHandleGeneral(
 
         if (NT_SUCCESS(status))
         {
-            NTSTATUS exitStatus = STATUS_PENDING;
+            BOOL isTerminated = FALSE;
```
dmex: NT doesn't use the Win32 BOOL type.

ProcessHacker/hndlprp.c
```diff
@@ -1111,10 +1111,34 @@ VOID PhpUpdateHandleGeneral(
                 PhDereferenceObject(name);
             }
 
+            NtQueryInformationThread(dupHandle,
+                ThreadIsTerminated,
+                &isTerminated,
+                sizeof(BOOL),
```
dmex: sizeof(ULONG)

Biswa96: Wouldn't it be easier to use `sizeof isTerminated`? One has to just change the variable declaration.

diversenok: @Biswa96, For some reason, this approach is hardly ever used throughout the codebase. Don't know why, though.

diversenok:
Done.

ProcessHacker/hndlprp.c
dmex: @Biswa96 @diversenok It makes it easier for anyone (even without experience) to review the calls 
 and for consistency.

PR #494, opened 2019-10-29
Make several functions from ntrtl.h take const strings

zeffy:
I went through all the functions in `ntrtl.h` that took non-`const` string parameters (like `PCHAR`) that were marked as `_In_` and checked the documentation on MSDN (or reactOS if undocumented) to see if the parameters were `const` or not, and updated the definitions accordingly. I didn't touch any functions that I couldn't find sufficient documentation on.

phlib and ProcessHacker compile cleanly with no warnings caused from these changes.

### DIFF
--- a/phnt/include/ntrtl.h
+++ b/phnt/include/ntrtl.h
@@ -1248,7 +1248,7 @@ RtlWakeAddressSingle(
 #ifndef PHNT_NO_INLINE_INIT_STRING
 FORCEINLINE VOID RtlInitString(
     _Out_ PSTRING DestinationString,
-    _In_opt_ PSTR SourceString
+    _In_opt_ PCSTR SourceString
     )
 {
     if (SourceString)
@@ -1256,7 +1256,7 @@ FORCEINLINE VOID RtlInitString(
     else
         DestinationString->MaximumLength = DestinationString->Length = 0;
 
-    DestinationString->Buffer = SourceString;
+    DestinationString->Buffer = (PCHAR)SourceString;
 }
 #else
 NTSYSAPI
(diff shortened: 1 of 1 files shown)


dmex:
> documentation on MSDN (or reactOS if undocumented)

Do you know why strings should be const without also using const with func(void*) or func(struct*)?

zeffy:
I was running into issues using e.g. `RtlInitUnicodeString` with a `const wchar_t[]` literal, because the compiler insisted it should be non-const due to the function definition. 

So strings were my main focus here, but that said I don't believe other parameters for the functions I changed can be made const (at least according to the documentation I found), but yes there are probably more functions with parameters that could be made const.

dmex:
> running into issues using e.g. RtlInitUnicodeString with a const wchar_t[] literal

What type are you passing for the string L"" or a c_str()?

> parameters that could be made const.

You have to cast everywhere which overrides the compiler and becomes painful?
https://github.com/processhacker/processhacker/blob/7853438778ab6bd573c7a11996b536867213c37b/phnt/include/ntrtl.h#L1421-L1431

zeffy:
>What type are you passing for the string L"" or a c_str()?

L""

>You have to cast everywhere which overrides the compiler and becomes painful?

That is exactly how it is done in both the WRK and [reactOS](https://github.com/reactos/reactos/blob/master/sdk/tools/mkhive/rtl.c#L45). I believe the logic is that it only uses the cast to fill a struct field and return it back to the caller who knows if the data is actually const or not. It is not casting away `const` to write to memory that it shouldn't be.

dmex:
@zeffy 

I've merged the changes. They'll also be merged into the PHNT repository 👍

PR #529, opened 2019-12-22
Fix analyzer warnings from WindowExplorer plugin.

poizan42:
I'm not entirely sure if some of the checks should be asserts, but this is how it seems to be done in other parts of the code.

### DIFF
--- a/plugins/WindowExplorer/prpsh.h
+++ b/plugins/WindowExplorer/prpsh.h
@@ -81,7 +81,9 @@ PPV_PROPPAGECONTEXT PvCreatePropPageContextEx(
     _In_opt_ PVOID Context
     );
 
-BOOLEAN PvPropPageDlgProcHeader(
+BOOLEAN
+_Success_(return)
+PvPropPageDlgProcHeader(
     _In_ HWND hwndDlg,
     _In_ UINT uMsg,
     _In_ LPARAM lParam,
@@ -102,7 +104,9 @@ VOID PvDoPropPageLayout(
     _In_ HWND hwnd
     );
 
-FORCEINLINE BOOLEAN PvPropPageDlgProcHeader(
+FORCEINLINE
+_Success_(return)
+BOOLEAN PvPropPageDlgProcHeader(
     _In_ HWND hwndDlg,
     _In_ UINT uMsg,
     _In_ LPARAM lParam,
--- a/plugins/WindowExplorer/wndprp.c
+++ b/plugins/WindowExplorer/wndprp.c
@@ -272,7 +272,7 @@ NTSTATUS WepPropertiesThreadStart(
 
     PhInitializeAutoPool(&autoPool);
 
-    if (propContext = HdCreatePropContext(PhaFormatString(L"Window %Ix", context->WindowHandle)->Buffer))
+    if (propContext = HdCreatePropContext(PhaFormatString(L"Window %Ix", (ULONG_PTR)context->WindowHandle)->Buffer))
     {
         PPV_PROPPAGECONTEXT newPage;
 
@@ -314,6 +314,9 @@ BOOLEAN NTAPI EnumGenericModulesCallback(
     )
 {
     PWINDOW_PROPERTIES_CONTEXT context = Context;
+
+    if (!context)
+        return FALSE;
 
     PhLoadModuleSymbolProvider(context->SymbolProvider, Module->FileName->Buffer,
         (ULONG64)Module->BaseAddress, Module->Size);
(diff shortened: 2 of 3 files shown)


plugins/WindowExplorer/prpsh.h
```diff
@@ -81,7 +81,7 @@ PPV_PROPPAGECONTEXT PvCreatePropPageContextEx(
     _In_opt_ PVOID Context
     );
 
-BOOLEAN PvPropPageDlgProcHeader(
+BOOLEAN _Success_(return) PvPropPageDlgProcHeader(
```
dmex: Multiple statements need to be multi-line. For example:

FORCEINLINE
_Success_(return)
BOOLEAN

plugins/WindowExplorer/prpsh.h
```diff
@@ -102,7 +102,7 @@ VOID PvDoPropPageLayout(
     _In_ HWND hwnd
     );
 
-FORCEINLINE BOOLEAN PvPropPageDlgProcHeader(
+FORCEINLINE _Success_(return) BOOLEAN PvPropPageDlgProcHeader(
```
dmex: Same here

plugins/WindowExplorer/wndprp.c
```diff
@@ -1030,7 +1033,7 @@ BOOL CALLBACK EnumPropsExCallback(
     {
         PPH_STRING propName;
 
-        propName = PhFormatString(L"#%lu", (ULONG_PTR)lpszString);
+        propName = PhFormatString(L"#%lu", (USHORT)lpszString);
```
dmex: The format specifier has %lu for unsigned long when you've cased to USHORT

poizan42: True, but we just checked that it fit in a short. The format specifier should probably be changed (?)

plugins/WindowExplorer/wndtree.c
```diff
@@ -365,12 +368,18 @@ BOOLEAN NTAPI WepWindowTreeNewCallback(
...
+        return FALSE;
+
     switch (Message)
     {
     case TreeNewGetChildren:
         {
             PPH_TREENEW_GET_CHILDREN getChildren = Parameter1;
 
+            if (!getChildren)
+                return FALSE;
```
dmex: Please use break here instead of returning inside the treeview message handler.

plugins/WindowExplorer/wndtree.c
```diff
@@ -420,6 +429,9 @@ BOOLEAN NTAPI WepWindowTreeNewCallback(
         {
             PPH_TREENEW_IS_LEAF isLeaf = Parameter1;
 
+            if (!isLeaf)
+                return FALSE;
```
dmex: break

plugins/WindowExplorer/wndtree.c
```diff
@@ -432,6 +444,9 @@ BOOLEAN NTAPI WepWindowTreeNewCallback(
         {
             PPH_TREENEW_GET_CELL_TEXT getCellText = Parameter1;
 
+            if (!getCellText)
+                return FALSE;
```
dmex: break

plugins/WindowExplorer/wndtree.c
```diff
@@ -462,6 +477,9 @@ BOOLEAN NTAPI WepWindowTreeNewCallback(
         {
             PPH_TREENEW_GET_NODE_COLOR getNodeColor = Parameter1;
 
+            if (!getNodeColor)
+                return FALSE;
```
dmex: break

plugins/WindowExplorer/wndtree.c
```diff
@@ -481,6 +499,9 @@ BOOLEAN NTAPI WepWindowTreeNewCallback(
         {
             PPH_TREENEW_KEY_EVENT keyEvent = Parameter1;
 
+            if (!keyEvent)
+                return FALSE;
```
dmex: break

dmex:
>  not entirely sure if some of the checks should be asserts

No. msvc is very dumb when it comes to SAL checks. Some treenew callbacks use NULL but others don't and the msvc analyzer can't tell the difference so you end up with warnings for one or the other.

We should be using "`_In_opt_ _When_(Message == TreeNewGetChildren, _In_))`" so that _In_opt_ would be used for everything except specific messages or a similar SAL annotation (like the win32 API does in these cases) but these annotations have never worked or been fixed.

This is why I use the "if (null) break;" since most will be optimized away and it shuts up the warning 😄

poizan42:
Should be fixed

dmex:
> True, but we just checked that it fit in a short. The format specifier should probably be changed (?)

I mean if you cast to ushort in the format parameter then the format specifier should be updated just for consistency.

> Should be fixed

Thanks!

PR #568, opened 2020-03-06
phlib/baseup: add C alternative in PhFindCharInStringRef and PhFindLastCharInStringRef

Biswa96:


Biswa96:
If you don't want to link msvcrt then these may be replaced with do while loop.

diversenok:
By the way, I am not entirely sure how it works in C, but I see that ntdll exports some functions that are usually attributed to C runtime library, like `wcsrchr` or `strlen`. It seems that we can use them without bringing any dependencies. Am I right?

Biswa96:
I want to skip any ntdll linking, one of the reason is that PH is already becomes Windows Defender's favorite one. A do while loop or msvcrt linking may not create any issue. Though after all last decision is up to dmex.

Biswa96:
@dmex any comment?

dmex:
> If you don't want to link msvcrt then these may be replaced with do while loop.

Is msvcrt linking an issue? @Biswa96 



> I am not entirely sure how it works in C, but I see that ntdll exports some functions that are usually attributed to C runtime library, like wcsrchr or strlen. It seems that we can use them without bringing any dependencies.

If you're using /NODEFAULTLIB then you can link against ntdllp.lib (not to be confused with ntdll.lib) for the CRT @diversenok

PR #632, opened 2020-06-21
ARM64 CPU info improvements

ge0rdi:
* **Read processor name from registry**  
Windows Task Manager does it that way.
* **Better speed readings for heterogeneous CPUs**
This way more accurate information about CPU speed is displayed on big.LITTLE ARM devices.
Actual and base CPU speed now matches the one shown in standard task manager (and imo the correct one).


dmex:
Thanks for the PR, looks good 👍

ge0rdi:
@dmex 
Thank you.

But what was wrong with `ARM64: Allow deployment to remote machine in debug config` commit?
I've enabled deployment for ARM64 Debug configurations for easier remote debugging.
With deployment enabled VS can automatically transfer new binaries to remote machine.
So there is no need to manually copy them when starting debug session.

dmex:
> Allow deployment to remote machine in debug config commit?

You would need to add it to all configurations not just the arm64 config to be accepted.

The debugging deployment configuration is generally different for each machine and since those changes were only for arm64, I figured it was unintentionally committed.

ge0rdi:
I just though it was not necessary for other than ARM64 configs, since nobody did it yet. Most people has x64 machines, so it is much easier to debug locally. But for ARM the only real option is remote debugging righ now.

I will submit another PR then.

PR #635, opened 2020-06-25
Build: ARM64 support for `-sdk` switch

ge0rdi:


dmex:
There are some issues with this one.

* It's bundling the ARM files with the 32bit and 64bit files but ARM is not compatible with either of those configurations.
* This PR also deleted the CustomBuildTool executable which is required for all the builds. Since it's a .NET core application the release binary is compiled by right-clicking the solution and selecting Publish which compiles the executable.

I need to make major changes to the build tools so that the ARM files are located separately from the 32bit/64bit files otherwise ARM users will run into a lot of issues.

ge0rdi:
> This PR also deleted the CustomBuildTool executable which is required for all the builds.

That's weird, because I can see it is still present.
Just few KB smaller (for whatever reason).

```
diff --git a/tools/CustomBuildTool/bin/Release/CustomBuildTool.exe b/tools/CustomBuildTool/bin/Release/CustomBuildTool.exe
index f9ba7a86..6b393293 100644
Binary files a/tools/CustomBuildTool/bin/Release/CustomBuildTool.exe and b/tools/CustomBuildTool/bin/Release/CustomBuildTool.exe differ
```

> Since it's a .NET core application the release binary is compiled by right-clicking the solution and selecting Publish which compiles the executable.

That's exactly what I did.

> It's bundling the ARM files with the 32bit and 64bit files but ARM is not compatible with either of those configurations.

What is the problem exactly?
It only copies few ARM64 files to separate ARM64 folders. Both x86/x64 and ARM64 build work well with those changes.

> I'll be adding this soon with some other required updates

That would be really cool.
Since current state is not exactly ARM64 friendly.

dmex:
> That's weird, because I can see it is still present.

Ah, yeah it does. The Github redesign is showing the entry as Red like it was removed when its actually changed.

> What is the problem exactly?

The build tools need to separate the ARM files from the 32bit/64bit files otherwise they'll be zipped/distributed with the (32bit) installer for 32/64bit Windows and there are issues with the CustomSetupTool and CustomSignTool projects the build tools depend on not currently having ARM versions breaking the driver and other parts of the build tool.

> Since current state is not exactly ARM64 friendly.

The move to 64bit took a few months and the move to ARM will probably take about the same time - I've already made most changes but testing some others is a bit slow because I don't currently have ARM hardware :/

ge0rdi:
> The build tools need to separate the ARM files from the 32bit/64bit files otherwise they'll be zipped/distributed with the (32bit) installer for 32/64bit Windows and there are issues with the CustomSetupTool and CustomSignTool projects the build tools depend on not currently having ARM versions breaking the driver and other parts of the build tool.

I've tried to build whole package (using `build_release.cmd`) and the only obsevable presence of ARM64 files in output files were libraries in `processhacker-build-sdk.zip`. They are placed in separate folder, so I don't think they could interfere with anything.
Other output files didn't contain any traces of arm64 related binaries.

> I've already made most changes but testing some others is a bit slow because I don't currently have ARM hardware :/

I guess the build process can be adjusted even without real HW (as build will run on x86/x64).
But surely testing with real HW makes things a lot easier.

Anyways, thanks for all the effort (it is really great to have real process inspection tool on ARM64 device :) ).

PR #636, opened 2020-06-25
Enable deployment for project configurations

ge0rdi:
It will allow Visual Studio to configure automatic deployment of binaries to remote machine.
So remote debugging/testing can be much easier.

@dmex 
Is it OK this way?
Should I enable deployment also for Plugins?


dmex:
> Should I enable deployment also for Plugins?

Is it possible to enable this from the .props file?

ge0rdi:
> Is it possible to enable this from the .props file?

Don't think so.
Deployment is enabled in solution for each project separately. Props are included into project files.

PR #666, opened 2020-08-25
Add support for Shadow Stack (CET) related change

TheEragon:
This PR adds new column under processes and modules which displays which processes and modules are CET capable. I wasn't sure how to name the column since under mitigations this is called Shadow Stack (which I find too long 😄) but Microsoft internally calls this CET, based on the DllCharaceristicsEx.

I have also added new feature to peview to display EH continuation metadata, which are also used by CET. There is a slight hack here, Windows 20H1 SDK defines IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT as 0x00200000 but if you look at 21H1 SDK there it's 0x00400000 with a note that 0x00200000 is reserved since it was used in 20H1, my guess is they screwed up and had to change the flag, so I manually check both values.


phlib/mapimg.c
```diff
@@ -2371,3 +2469,98 @@ BOOLEAN PhGetMappedImageDebugEntryByType(
...
+    PIMAGE_LOAD_CONFIG_DIRECTORY32 config32;
+
+    if (!NT_SUCCESS(status = PhGetMappedImageLoadConfig32(MappedImage, &config32)))
+        return status;
+
+    // Not every load configuration contains eh continuation
+    if (!RTL_CONTAINS_FIELD(config32, config32->Size, GuardEHContinuationCount))
+        return STATUS_INVALID_VIEW_SIZE;
+
+    EhContConfig->EhContTable = PhMappedImageRvaToVa(MappedImage, (ULONG)(config32->GuardEHContinuationTable - MappedImage->NtHeaders->OptionalHeader.ImageBase), NULL);
```
dmex, line 2488: This needs to be MappedImage->NtHeaders32 for the 32bit headers instead of MappedImage->NtHeaders 

Optionally that statement could also be: PtrToUlong(PTR_SUB_OFFSET(config32->GuardEHContinuationTable, MappedImage->NtHeaders32->OptionalHeader.ImageBase)) but that's entirely personal style and optional.

phlib/mapimg.c
```diff
@@ -2371,3 +2469,98 @@ BOOLEAN PhGetMappedImageDebugEntryByType(
...
+    EhContConfig->NumberOfEhContEntries = config32->GuardEHContinuationCount;
+
+    // taken from from nt!RtlGuardRestoreContext
+    EhContConfig->EntrySize = ((config32->GuardFlags & IMAGE_GUARD_CF_FUNCTION_TABLE_SIZE_MASK) >> IMAGE_GUARD_CF_FUNCTION_TABLE_SIZE_SHIFT) + 4;
```
dmex, line 2492: RtlGuardRestoreContext uses `sizeof(ULONG)` instead of `+4` since the table entries are ULONGs

phlib/mapimg.c
```diff
@@ -2371,3 +2469,98 @@ BOOLEAN PhGetMappedImageDebugEntryByType(
...
+
+    // Not every load configuration contains eh continuation
+    if (!RTL_CONTAINS_FIELD(config64, config64->Size, GuardEHContinuationCount))
+        return STATUS_INVALID_VIEW_SIZE;
+
+    EhContConfig->EhContTable = PhMappedImageRvaToVa(MappedImage, (ULONG)(config64->GuardEHContinuationTable - MappedImage->NtHeaders->OptionalHeader.ImageBase), NULL);
+    EhContConfig->NumberOfEhContEntries = config64->GuardEHContinuationCount;
+
+    // taken from from nt!RtlGuardRestoreContext
+    EhContConfig->EntrySize = ((config64->GuardFlags & IMAGE_GUARD_CF_FUNCTION_TABLE_SIZE_MASK) >> IMAGE_GUARD_CF_FUNCTION_TABLE_SIZE_SHIFT) + 4;
```
dmex, line 2532: sizeof(ULONG) vs +4

phlib/mapimg.c
```diff
@@ -646,6 +646,104 @@ NTSTATUS PhUnloadRemoteMappedImage(
...
+            sizeof(IMAGE_DEBUG_DIRECTORY),
+            NULL
+        );
+
+        if (!NT_SUCCESS(status))
+            return FALSE;
+
+        if (currentDebugEntry.Type == Type)
+        {
+            *DataBuffer = PhAllocate(currentDebugEntry.SizeOfData);
```
dmex, line 721: This probably shouldn't touch the callers buffer unless the function succeeds. Can this be changed to a local variable?

phlib/mapimg.c
dmex, line 2488: NtHeaders32 instead of NtHeaders

phlib/mapimg.c
dmex, line 2492: sizeof(ULONG)

phlib/mapimg.c
dmex, line 2532: same as previous

dmex (changes requested):

phlib/mapimg.c
dmex: The main reason for this change is so that the +4 isn't some magical unexplained value when we know its the sizeof table entry 👍🏿

dmex:
I don't see any real problems with this PR just some minor style issues. I'm quite impressed you've used the correct macros and functions properly 👍🏿 

I can merge it without requiring those changes if you'd prefer me to make the changes instead?

TheEragon:
> I don't see any real problems with this PR just some minor style issues. I'm quite impressed you've used the correct macros and functions properly 👍🏿
> 
> I can merge it without requiring those changes if you'd prefer me to make the changes instead?

I'll fix it in a minute, you have valid points. Do you want me to do --fixup commit or just fix those that I already pushed and force push them?

phlib/mapimg.c
TheEragon: Yes, I wanted to use sizeof(ULONG) but forgot about, it seems some modules compiled with /ehcont have extra 0 added to each entry. I think those are flags, but it was always 0 so I didn't bother adding it to the table in peview yet.

dmex:
> just fix those that I already pushed and force push them?

It's whatever you prefer will be fine. Github will squash the commits into a single merge

phlib/mapimg.c
TheEragon: Nice catch, EHCont doesn't seem to be supported on x86, so there is no way to test this branch, I added it just in case MS adds support for in the future, which I doubt they will since exceptions work in a different way on x86.

TheEragon:
> > just fix those that I already pushed and force push them?
> 
> It's whatever you prefer will be fine. Github will squash the commits into a single merge

Ah...good to know, I'll make 2 PR then next time :smile:

dmex:
> I'll make 2 PR then next time 😄

It wouldn't be a problem if you did. I sometimes forget to commit something and have to merge > 5 commits so I can't complain if anyone else uses multiple PR/commits too 😃

TheEragon:
> > I'll make 2 PR then next time 😄
> 
> It wouldn't be a problem if you did. I sometimes forget to commit something and have to merge > 5 commits so I can't complain if anyone else uses multiple PR/commits too 😃

Changes were just pushed, I hope I fixed everything you pointed out.

phlib/mapimg.c
dmex: Those modules might be compiled with LLVM or an earlier version of the MSVC linker which uses 0 for determining the end of the table?

dmex (approved):

phlib/mapimg.c
TheEragon: No, I used same compiler (VS 16.7.2) on different projects, but one was EXE (size of entry was 4, so just RVA) and one was DLL (size 5).

dmex: The last comment was incorrect and I removed it. The change in size might be the actual flag for determining if the module is an EXE vs a DLL.

MSVC also currently uses two different versions of the same compiler and can be seen in the RICH metadata (ProdID) tab in PEViewer so it could also be an older version creating the size difference... I guess we'll know eventually what causes it.

dmex:
> Changes were just pushed, I hope I fixed everything you pointed out.

Looks good. The nightly build has been updated with the changes.

TheEragon:
> > Changes were just pushed, I hope I fixed everything you pointed out.
> 
> Looks good. The nightly build has been updated with the changes.

Yes I noticed, everything seems to work as expected on my end. Thank you for the quick build!

Biswa96:
About the different const. values. Can this be added just for readability?

```c
#ifndef IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT_20H1
#define IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT_20H1 0x00200000
#endif

#ifndef IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT_21H1
#define IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT_21H1 0x00400000
#endif
```

Or whatever the code name is _VB, _MN, _FE etc in sdkddkver.h in WinSDK.

TheEragon:
I'd say you don't need to guard them with `#ifndef` since MS won't ever add those to winnt.h

dmex:
> MS won't ever add those to winnt.h

Microsoft already defined the values in the current 20H1 (19041) SDK line 18701 winnt.h:
`#define IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT      0x00200000 // Module contains EH continuation target information`

> About the different const. values. Can this be added just for readability?

I'm not sure if the name was redefined with 0x00400000 which would break querying the table on 20H1 but I doubt there are breaking SDK changes/renames.

Biswa96:
@TheEragon said MS won't ever use the _20H1 or _VB suffixes. The following is in WinSDK Insider build 20190:

```c
#define IMAGE_GUARD_RETPOLINE_PRESENT                  0x00100000 // Module was built with retpoline support
// DO_NOT_USE                                          0x00200000 // Was EHCont flag on VB (20H1)
#define IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT      0x00400000 // Module contains EH continuation target information
#define IMAGE_GUARD_XFG_ENABLED                        0x00800000 // Module was built with xfg
```

TheEragon:
> I'm not sure if the name was redefined with 0x00400000 which would break querying the table on 20H1 but I doubt there are breaking SDK changes/renames.

Actually they did, if you check winnt.h from 21H1 there is this
```c++
// DO_NOT_USE                                          0x00200000 // Was EHCont flag on VB (20H1)
#define IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT      0x00400000 // Module contains EH continuation target information```

Also current linker in MSVC 16.7.2 is also using 0x00400000, I think 0x00200000 was a fluke.

Biswa96:
That's why I have suggested to use the _20H1 or _VB version/codename suffixes.

dmex:
> use the _20H1 or _VB version/codename suffixes.

We did originally use the codenames but they were too unreliable/confusing for versioning and build testing. We have to stick with the version (20H1/20H2).

> Actually they did

Great. So we won't ever be able to use the SDK definition because the solutions automatically select whatever latest SDK you've installed and compiling against the SDK for the define will break support for either older versions of Windows 10 or break support for newer builds of Windows 10.

dmex:
@TheEragon 

I would use add the structure version to the name since V2 will be used in future versions of Win10 not just 21H1:

IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT_V1
IMAGE_GUARD_EH_CONTINUATION_TABLE_PRESENT_V2

TheEragon:
I'd go with _20H1 and _21H1 since in sdk it says `// Was EHCont flag on VB (20H1)`, but you're the boss here :smile:

Do you want me to do it or @Biswa96 will you do it?

PR #669, opened 2020-08-26
Support for XFG

TheEragon:
XFG is the successor of CFG, currently supported from Windows 21H1.

### DIFF
--- a/ProcessHacker/include/modprv.h
+++ b/ProcessHacker/include/modprv.h
@@ -37,6 +37,7 @@ typedef struct _PH_MODULE_ITEM
     USHORT ImageCharacteristics;
     USHORT ImageDllCharacteristics;
     ULONG ImageDllCharacteristicsEx;
+    ULONG GuardFlags;
 
     LARGE_INTEGER LoadTime;
     LARGE_INTEGER FileLastWriteTime;
--- a/ProcessHacker/modlist.c
+++ b/ProcessHacker/modlist.c
@@ -31,6 +31,9 @@
 #include <modprv.h>
 #include <phsettings.h>
 
+// remove once IMAGE_GUARD_XFG_ENABLED is defined (TheEragon)
+#include <mapimg.h>
+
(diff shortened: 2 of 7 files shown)


TheEragon:
Current implementation only shows in modules list which modules have XFG or CFG support.
[Screenshot 2020-08-26 142737
](url)

I'm not sure what to do with process list since that just shows either "CF Guard" or nothing. Here are couple of ideas I had

1. Display "XF Guard" if main executable supports XFG
2. Display "XF Guard" if all loaded modules support XFG
3. Rename "CF Guard" to something generic like "On"
4. Leave it as is, since you can always open the process and check modules

I was inclined to go with option 1) but now I'm thinking that 4) sounds the best, what do you think? Or do you have some other idea how to display it?

Also, do we want to keep the name of the column?

dmex:
> Display "XF Guard" if all loaded modules support XFG

dumpbin shows an X next to functions using XFG while showing the CFG function table so the module might always be using XFG for every function protected by CFG?

>  I'm thinking that 4) sounds the best

The process can have modules using both CFG and XFG, so showing XFG for the process might not be accurate in those cases so 4) would likely be the best candidate.

> do we want to keep the name of the column?

Yeah, I guess using CFG would be the best since XFG is still CFG.

I've also discovered XFG can also be used to check if reverse engineered function prototypes for undocumented functions are correct by hashing the prototype and checking against the XFG hash. Since the number of parameters and types are limited you could also easily brute force prototypes for undocumented functions?

dmex (approved):

Biswa96:
Easter egg in WinSDK 20215:

```c
#ifndef DECLSPEC_XFGVIRT
#if _CONTROL_FLOW_GUARD_XFG
#define DECLSPEC_XFGVIRT(base, func) __declspec(xfg_virtual(base, func))
#else
#define DECLSPEC_XFGVIRT(base, func)
#endif
#endif
```

PR #671, opened 2020-08-27
Add support for new flags in PROCESS_MITIGATION_USER_SHADOW_STACK_POLICY

TheEragon:


dmex (approved):

TheEragon:
I modified the strings to be consistent and start with a lower letter after : in long descriptions.

dmex (approved):

dmex:
I wasn't able to merge this right away as it was blocked on this commit: https://github.com/processhacker/processhacker/commit/aa7a04476982425cd654dfd042bee8bbaf1a877c

It's now been merged 👍🏿 

------

It was unrelated to the PR but there was an internal compiler and linker error building the source that required investigation. The issue was because LTGC wasn't defined and defaulted to LTGC:Incremental which produces corrupted PDBs and several random internal compiler errors when combined with the CET, XFG, EHCONT and some other options:

image

So if you have issues compiling code with CET, XFG, EHCONT etc... make sure LTGC has been defined otherwise MSVC defaults to LTGC:Incremental and produces corrupted PDBs and/or unexplained internal compiler/linker errors.

TheEragon:
> So if you have issues compiling code with CET, XFG, EHCONT etc... make sure LTGC has been defined otherwise MSVC defaults to LTGC:Incremental and produces corrupted PDBs and/or unexplained internal compiler/linker errors.

Interesting, thanks for the info 👍. By the way, XFG isn't available in VS yet, at least not publicly, you have to use switch /guard:xfg to turn it on.

dmex:
> use switch /guard:xfg

Yep, also ehcont will be delayed until the next SDK update since the current CRT libs don't support it properly yet.

It's similar with the /brepro switch except Microsoft created a webpage stating all unofficial builds of Process Hacker source-code are malicious and added signatures to Defender enforcing it. So if builds are reproducible then both official and unofficial would be byte identical possibly creating some issues... I'm waiting for a response to an email before enabling brepro.

TheEragon:
> Yep, also ehcont will be delayed until the next SDK update since the current CRT libs don't support it properly yet...
 
Yes and a fact that 20H1 kernel checks for flag 0x00200000 while the newest VS sets 0x00400000 😄 so even if you set the flag and generate the data kernel ignores them. But who know maybe they'll release a fixed SDK with 20H2

> /brepro is another issue which is also delayed.

Got more information on this somewhere?

dmex:
> Got more information on this somewhere?

* Random compiler/build errors caused by LTGC:Incremental when combined with /BREPRO. I found disabling /MP fixed the initial few errors but it caused a much longer compile time for the project. The actual issue was LTGC being undefined in the solution and defaulting to LTGC:Incremental instead of LTGC which when combined with /MP (build with multiple processes) produces invalid executables when used with /BREPRO.

* Antivirus false positives... Antivirus doesn't support the image timestamp field as the deterministic hash and incorrectly determine the binary as malicious because of the invalid timestamp. 


* Stripping PDB paths using PDBALTPATH

dmex:
@TheEragon 

EHCONT has now been enabled and the nightly build was updated successfully 👍

TheEragon:
@dmex  Nice, have you switched the SDK to 21H1 or have they fixed 20H2?

dmex:
@TheEragon 

I fixed the buildtools ignoring the status code which was successfully compiling but being ignored. When the SDK updates or you install a newer version it will now also automatically switch to the newer version.

VS is compiling fine and generating the tables for the nightly executables so it should be working on 21H1 but other than that there isn't much else I can do if Windows isn't working properly.

dmex:
@TheEragon 

EHCONT seems to be completely broken in Visual Studio at the moment. No images are compiling with GuardEHContinuationTable despite the /guard:ehcont flags in the solutions?

TheEragon:
> @TheEragon
> 
> EHCONT seems to be completely broken in Visual Studio at the moment. No images are compiling with GuardEHContinuationTable despite the /guard:ehcont flags in the solutions?

That is strange, did it break after switching to VS16.9.4?

dmex:
@TheEragon 

Yeah, looks like it. There's a patch coming in one of the next updates: 
https://developercommunity.visualstudio.com/t/Enabling-EH-Continuation-Metadata-doesn/1400623

TheEragon:
@dmex 

Just great, I wonder how much stuff gets broken with each new VS update, it feels like they fix 5 things and break another 6 sometimes.

PR #692, opened 2020-09-21
Align & improve updater wordings 

Eagle3386:
Aligns the check button's caption with common best practices (1 word on button) & improves the instruction's texts while check is running & the reported sizes/speed in more modern units.

Please review & suggest edits where needed.

### DIFF
--- a/plugins/Updater/page1.c
+++ b/plugins/Updater/page1.c
@@ -24,7 +24,7 @@
 
 static TASKDIALOG_BUTTON TaskDialogButtonArray[] =
 {
-    { IDOK, L"Check for updates" }
+    { IDOK, L"Check" }
 };
 
 HRESULT CALLBACK CheckForUpdatesCallbackProc(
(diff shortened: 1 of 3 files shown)


dmex (approved):

plugins/Updater/page1.c
```diff
@@ -24,7 +24,7 @@
 
 static TASKDIALOG_BUTTON TaskDialogButtonArray[] =
 {
-    { IDOK, L"Check for updates" }
```
dmex, line 27: This was probably used more for space filler 😄 

There isn't much content on the dialog and Check on its own feels too short. you can get away with extra text for the button in this case.

dmex (changes requested):

plugins/Updater/page1.c
Eagle3386: Sure, since that long wording was my initial reason for refactoring and I rewrote the instruction's text, too, can we shorten this one, please?
As an argument: why not un-comment the longer content's text below instead of the short one? 😄

dmex: > why not un-comment the longer text below

I didn't think clicking the button needed the longer instruction text 😛

PR #740, opened 2020-11-20
Add group SID in Token tab

revengsh:
Hi!
I'd like to propose this little improvement. I found that browsing groups in the Token tab without having their SID was a bit unconvenient.
So i added a translation of the selected group name to its SID, at the bottom of the ListView.


dmex (approved):

dmex:
I've added the SIDs as an extra column for sorting and eventually search 👍

image

dmex (changes requested):
Can you copy my changes to your branch and update it first so I can merge the PR? 

Gihub didn't update the coauthor in the commit (ac61b227182c4f8de3f7cb84968f14271b7edcf4). I need to merge authors instead of adding them as coauthors until the issue is fixed.

dmex (approved):

PR #751, opened 2020-12-12
Implement Image Coherency

jxy-s:
# Image Coherency
This PR implements the Image Coherency described in https://github.com/processhacker/processhacker/issues/744. Please see that issue for a detailed rundown of the changes.

## Summary
In response to research on process tampering (Hollowing, Doppelganing, Herpaderping). This is an enhancement to Process Hacker which inspects the coherency between an executing process and the image on disk.

For context on Herpaderping please see: https://jxy-s.github.io/herpaderping/

This enhancement adds an API to `phlib` which inspects the image on disk and the image mapped into the process. This calculation represents the similarity between the image on disk and the mapped image. This is displayed in the process tree by "noting" the process' which have "Low Image Coherency". In addition, enhances the process table with a column displaying the "Image Coherency".

### DIFF
--- a/phlib/imgcoherency.c
+++ b/phlib/imgcoherency.c
@@ -1,0 +1,833 @@
+/*
+ * Process Hacker -
+ *   Image Coherency
+ *
+ * Copyright (C) 2020 jxy-s
+ *
+ * This file is part of Process Hacker.
+ *
+ * Process Hacker is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Process Hacker is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Process Hacker.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <ph.h>
+#include <mapimg.h>
+
+/**
+* Process image coherency context, used during the calculation of the process
+* image coherency.
+*/
+typedef struct _PH_PROCESS_IMAGE_COHERENCY_CONTEXT
+{
+    NTSTATUS MappedImageStatus;               /**< Status of initializing MappedImage */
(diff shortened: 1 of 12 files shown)


phlib/imgcoherency.c
```diff
@@ -0,0 +1,833 @@
...
+    PPH_PROCESS_IMAGE_COHERENCY_CONTEXT context;
+    PROCESS_BASIC_INFORMATION basicInfo;
+
+    //
+    // This is best-effort context creation, we don't fail if the mapping
+    // fails - the caller of the API may make decisions based on success of
+    // each
+    //
+
+    context = PhAllocateZero(sizeof(PH_PROCESS_IMAGE_COHERENCY_CONTEXT));
```
dmex, line 86: PhAllocate throws on allocation failure while PhAllocateSafe returns NULL

phlib/imgcoherency.c
```diff
@@ -0,0 +1,833 @@
...
+    // The RVA helper function uses a buffer we supply. For each RVA we will
+    // try to inspect a page - make one
+    //
+    // We only inspect 1 page for each for native code block for simplicity.
+    // Once the native PE is mapped parts the code are "fixed up" by the
+    // loader. It would take substantial work to "unmap" the remote image
+    // to something more coherent and would be time-complex. This gets us
+    // most of the way at the cost accuracy - but it's good enough.
+    //
+    buffer = PhAllocateZero(PAGE_SIZE);
```
dmex, line 291: Should this be `BYTE buffer[PAGE_SIZE]` instead of allocate?

dmex (commented):
LTGM

dmex (approved):

phlib/imgcoherency.c
dmex: PhAllocatePage might be more useful

valk0:
Great job!
How does that work with dlls?

dmex:
> How does that work with dlls?

They're identical and use the same PE layout.

valk0:
> > How does that work with dlls?
> 
> They're identical and use the same PE layout.

I meant like which dlls do you compare? Do you recursively check dlls that should be loaded based on the import table on the disk? How does that work with dlls loaded in runtime (LoadLibrary)

jxy-s:
> I meant like which dlls do you compare?

DLLs are not inspected in this implementation. However as @dmex stated, the same calculation may be applied to them. So it may be extended to do so.

PR #767, opened 2021-01-02
KPH hardening, protected domination check and caller verification

jxy-s:
# KPH Hardening
This PR implements two security features into KPH.
1. KPH will now do a domination check between the requesting process and the target process for protected calls, this ensures that a calling process is not given excessive access to protected processes.
2. KPH will verify the client connecting to it and on each protected call it will verify that the call into the kernel was made within the verified extents of the process. The client is not given access unless its image passes verification.


mrsshr:
after merged, can we get a new version of the authenticode signed kprocesshacker?
i'm sad

KProcessHacker/include/ntfill.h
```diff
@@ -332,4 +394,222 @@ PsReleaseProcessExitSynchronization(
...
+typedef
+NTSTATUS
+(NTAPI *PRTL_IMAGE_NT_HEADER_EX)(
+    _In_ ULONG Flags,
+    _In_ PVOID Base,
+    _In_ ULONG64 Size,
+    _Out_ PIMAGE_NT_HEADERS* OutHeaders
+    );
+
+typedef struct _IMAGE_DOS_HEADER {      // DOS .EXE header
```
dmex, line 413: Should we instead be using ntimage.h from the WDK for these structures?

dmex (approved):

dmex:
LGTM with or without ntimage.h

diversenok:
> KPH will now do a domination check between the requesting process and the target process for protected calls, this ensures that a calling process is not given excessive access to protected processes.

Since Process Hacker doesn't run as a protected process, does it mean it won't be able to inspect protected processes anymore?

Also, which values of `SecurityLevel` for KProcessHacker3 does this change affect?

dmex:
> does it mean it won't be able to inspect protected processes anymore?

You can still inspect protected processes but you won't be able to terminate. 

@jxy-s 

Administrative override for SecurityLevel should also be removed.

PR #790, opened 2021-02-04
Improve dbghelp.dll finding on ARM64

Mattiwatti:
`PhpSymbolProviderCompleteInitialization` currently deals with the x64 and x86 cases to build the path to load `dbghelp.dll`, but due to the `_WIN64` ifdef, the x64 path will be tried on ARM64 as well even though it has a separate `arm64` directory in the SDK. This PR fixes this.

---

Note that thread stacks are currently far from optimal on ARM64 regardless of which dbghelp.dll is loaded. On my Raspberry Pi 4 with Insider Preview 10.0.21296.0 I get the following stacktrace using the System32 dbghelp:
image
(note the missing bottom frame, as well as no symbols being present due to the missing `symsrv.dll`)

Using the Windows 10.19041.0 SDK dbghelp/symsrv we do get the bottom frame - in fact, dbghelp likes it so much it doesn't want to stop repeating it:
image
So both of these are pretty bad, but I still prefer the SDK dbghelp since at least it comes with correct symbol names.

I didn't make an issue for the above as I am getting equally bad stacktraces in WinDbg, so I assume this is a problem in dbghelp.dll and not Process Hacker.

### DIFF
--- a/phlib/symprv.c
+++ b/phlib/symprv.c
@@ -275,8 +275,10 @@ VOID PhpSymbolProviderCompleteInitialization(
         PPH_STRING dbghelpName;
         PPH_STRING symsrvName;
 
-#ifdef _WIN64
+#if defined(_M_AMD64)
         PhMoveReference(&winsdkPath, PhConcatStringRefZ(&winsdkPath->sr, L"\\Debuggers\\x64\\"));
+#elif defined(_M_ARM64)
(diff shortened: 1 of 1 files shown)


dmex (approved):

dmex:
LGTM

PR #804, opened 2021-02-16
Fix commit hash length & link About's hash to GitHub

Eagle3386:
@dmex I thought short commit hashes should be displayed equally correct & linking the version's last commit within the About dialog to its GitHub page (just like the Changelog listing of the Updater dialog does) would be a good idea - hence this PR.

### DIFF
--- a/ProcessHacker/about.c
+++ b/ProcessHacker/about.c
@@ -56,7 +56,7 @@ static INT_PTR CALLBACK PhpAboutDlgProc(
 
 #if (PHAPP_VERSION_REVISION != 0)
             appName = PhFormatString(
-                L"Process Hacker %lu.%lu.%lu (%hs)",
+                L"Process Hacker %lu.%lu.%lu (<a href=\"https://github.com/processhacker/processhacker/commit/%hs\">%hs</a>)",
                 PHAPP_VERSION_MAJOR,
                 PHAPP_VERSION_MINOR,
                 PHAPP_VERSION_REVISION,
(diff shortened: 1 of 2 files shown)


dmex (approved):

dmex:
> short commit hashes should be displayed equally correct

Makes sense. There was an issue with Github ages ago not displaying the commit page for some short hashes so I added the extra character as a workaround but it should have been fixed by now.

I've merged the PR, thanks for taking the time to make the changes 👍

dmex:
The link added an extra %hs but wasn't passing an extra commit string causing a crash when opening the about dialog. I fixed the the PR in another commit: ad73f6e082f2778764f872992eef15eb13d966e0

I've also updated the dialog to support links for the version text 👍

Eagle3386:
Sorry that I've missed that & thanks for the quick fix! 👍🏻

Eagle3386:
@dmex Just had a quick look after updating to your commit's version & found that the commit hash in both, about & update dialog still show 8 chars instead of 7. Did you include my changes from d388111?

dmex:
> Did you include my changes

Yep. I just had to test the latest .NET update before merging the recompiled buildtools binaries with your changes. The latest nightly already includes the updated hashes.

Eagle3386:
Just updated & saw the change. Thanks for letting me being part of this ride! 😍

You know, I love PH since my discovery of it after searching for a better Process Explorer back then & find it hard to accept that C++ still feels that much different to my daily C# writing that I can't commit more to your project. Really sorry about that. 🥺

dmex:
> find it hard to accept that C++ still feels that much different to my daily C#

PH was entirely C# before version 2.0... A lot of the functionality is based on those original versions and are still basically idential to how you would do something in C# with our own PH_STRING_BUILDER idential to StringBuilder, PH_LIST the same as List<>, hashlist as PH_HASHTABLE, arraylist as PH_ARRAY etc...

There are slight changes from C# because its native but practically everything is the same to the original v1.x builds in C# with the only difference now that its C.... I think you'll find the PH code is easier to use than anything you'll find anywhere else just because of how it was ported over to C from C# by wj32 and a very decent way of doing things.

PR #816, opened 2021-02-28
Improve name generation for process and thread handles

diversenok:
Since it is possible to determine image names for processes without snapshotting or opening handles, I decided to reorganize the logic behind `PhStdGetClientIdName` and `PhGetClientIdName`. The new implementation does not invoke additional snapshotting and does not suffer from not finding freshly created processes in the snapshots. I also adjusted the output to use thread names (when available) and indicate termination state.

image

As a bonus, now you can also search for programs that leak handles to terminated threads and processes:

image

### DIFF
--- a/phlib/hndlinfo.c
+++ b/phlib/hndlinfo.c
@@ -849,7 +849,8 @@ PPH_STRING PhStdGetClientIdNameEx(
     PPH_STRING threadName = NULL;
     PH_STRINGREF processNameRef;
     PH_STRINGREF threadNameRef;
-    HANDLE hProcess, hThread;
+    HANDLE processHandle;
+    HANDLE threadHandle;
     ULONG isProcessTerminated = FALSE;
     ULONG isThreadTerminated = FALSE;
 
@@ -879,7 +880,7 @@ PPH_STRING PhStdGetClientIdNameEx(
 
     // Check if the process is alive, but only if we didn't get its name
     if (!ProcessName && NT_SUCCESS(PhOpenProcess(
-        &hProcess,
+        &processHandle,
         SYNCHRONIZE,
         ClientId->UniqueProcess
         )))
@@ -887,10 +888,10 @@ PPH_STRING PhStdGetClientIdNameEx(
         LARGE_INTEGER timeout = { 0 };
 
         // Waiting with zero timout checks for termination
-        if (NtWaitForSingleObject(hProcess, FALSE, &timeout) == STATUS_WAIT_0)
+        if (NtWaitForSingleObject(processHandle, FALSE, &timeout) == STATUS_WAIT_0)
             isProcessTerminated = TRUE;
 
-        NtClose(hProcess);
+        NtClose(processHandle);
     }
 
     PhInitializeStringRef(&threadNameRef, L"unnamed thread");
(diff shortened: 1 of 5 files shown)


phlib/hndlinfo.c
```diff
@@ -836,95 +836,113 @@ _Callback_ PPH_STRING PhStdGetClientIdName(
...
+    if (!ProcessName && NT_SUCCESS(PhOpenProcess(
+        &hProcess,
+        SYNCHRONIZE,
+        ClientId->UniqueProcess
+        )))
+    {
+        LARGE_INTEGER timeout = { 0 };
 
-    PhAcquireQueuedLockShared(&cachedProcessesLock);
+        // Waiting with zero timout checks for termination
```
dmex, line 889: Shouldn't this be calling PhGetProcessBasicInformation(hProcess)? ExitStatus will always return STATUS_PENDING when the process hasn't terminated.

phlib/hndlinfo.c
```diff
@@ -836,95 +836,113 @@ _Callback_ PPH_STRING PhStdGetClientIdName(
...
+    _In_ PCLIENT_ID ClientId,
+    _In_opt_ PPH_STRING ProcessName
+    )
+{
+    PPH_STRING result;
+    PPH_STRING processName = NULL;
+    PPH_STRING threadName = NULL;
+    PH_STRINGREF processNameRef;
+    PH_STRINGREF threadNameRef;
+    HANDLE hProcess, hThread;
```
dmex, line 852: Don't use hProcess, hThread on the same line. It should use processHandle, threadHandle.

phlib/hndlinfo.c
diversenok: I thought about it, but what if a processes exists with STATUS_PENDING?

dmex (approved):

phlib/hndlinfo.c
dmex: It's mentioned in the documentation for GetExitCodeProcess that programs shouldn't ever return STATUS_PENDING since it can cause infinate loops but we don't use loops so it shouldn't be an issue?

If you think it should use NtWaitForSingleObject then it can remain there. I just wanted to double check if that was intended?

diversenok:
Also, hndlphp.c still uses `PhStdGetClientIdName`. Should I change it to `PhGetClientIdName`? I didn't do it because it is slightly unclear why it was this way in the first place.

phlib/hndlinfo.c
diversenok: I don't mind doing either way. I see NtWaitForSingleObject slightly more reliable when it comes to false negatives, but it requires opening the target for SYNCHRONIZE rather than PROCESS_QUERY_LIMITED_INFORMATION. I don't know which one is better, to be honest.

dmex: I guess we'll find out which way works better eventually 😋

dmex (approved):

dmex:
> I didn't do it because it is slightly unclear why it was this way in the first place.

There used to be more functions that utilized PhSetHandleClientIdFunction to handle the name lookup but I've removed most of those cases over the last few years. I haven't removed them entirely because I wasn't sure those changes would need to be reverted.

>  hndlprp.c still uses PhStdGetClientIdName. Should I change it to PhGetClientIdName

We can see how these changes go then in a few weeks update those calls?

diversenok:
Yep, sure

PR #817, opened 2021-02-28
Add a column that indicates impersonating threads

diversenok:
Sometimes you want to determine which threads of a process have associated tokens. Currently, you can check it via the context menu by clicking through them one-by-one. It is tedious and unreliable, mostly because you cannot tell if a thread impersonates a token you cannot open or doesn't have one. So, I added a new optional column for the thread view that should reliably indicate the impersonation state.

image


dmex (approved):

dmex:
LGTM 

Not sure if the column name should be 'Impersonation' or something descriptive but we can come up with something later if its needed 👍

diversenok:
Yeah, calling the column "Impersonation" might be more descriptive, though, slightly longer. Feel free to change it if you see that it fits. 👍

PR #903, opened 2021-05-27
Add support for kernel CET

TheEragon:
This change will add support for kernel CET on systems that support it.

Process tree
image

Process mitigations
image

Process modules
image

Change is pretty simple, but contains two potential problems:

1. Detecting System process by PID, I don't really like it and I'm welcome to ideas
2. Populating `UserShadowStack` in `PhGetProcessMitigationPolicy` with data retrived using `NtQuerySystemInformation`.

### DIFF
--- a/ProcessHacker/include/modprv.h
+++ b/ProcessHacker/include/modprv.h
@@ -77,7 +77,8 @@ typedef struct _PH_MODULE_PROVIDER
             BOOLEAN ControlFlowGuardEnabled : 1;
             BOOLEAN IsSubsystemProcess : 1;
             BOOLEAN CetEnabled : 1;
-            BOOLEAN Spare : 4;
+            BOOLEAN CetStrictModeEnabled : 1;
+            BOOLEAN Spare : 3;
         };
     };
     UCHAR ImageCoherencyScanLevel;
--- a/ProcessHacker/modprv.c
+++ b/ProcessHacker/modprv.c
@@ -170,10 +170,12 @@ PPH_MODULE_PROVIDER PhCreateModuleProvider(
         if (WindowsVersion >= WINDOWS_10_20H1)
         {
             BOOLEAN cetEnabled;
+            BOOLEAN cetStrictModeEnabled;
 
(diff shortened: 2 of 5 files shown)


dmex (approved):

dmex:
> Detecting System process by PID, I don't really like it

This is the correct method for detecting the system process by PID with the `SYSTEM_PROCESS_ID` macro but modifying the process wrappers to call system wrappers would probably be like mixing oil and water? 😋

> Populating UserShadowStack in PhGetProcessMitigationPolicy with data retrived using NtQuerySystemInformation.

The PhGetProcess* macros only handle the processinfo classes and native.c generally deals with the systeminfo classes. It's done this way primarily for abstraction and helps debugging changes to the Native API... In this specific case for SystemShadowStackInfo I think these functions should have their own wrappers and the caller determines what function to invoke since the places where they're used already have all that information and it'll keep those functions dedicated to those specific infoclasses.

It won't look as clean and tidy as having a single function that does both process and system and would require a few more changes but would give better abstration/more freedom for the caller to handle future changes to those classes.

I'm a bit surprised NtQueryInformationProcess doesn't work for the system process (which might be an oversight by MS?) you won't have to do this for other infoclasses so I wouldn't put much thought into having to do the above abstraction for other commits so just forget everything I mentioned here 😂

I've merged the PR and I'll update those wrappers 👍

TheEragon:
> I'm a bit surprised NtQueryInformationProcess doesn't work for the system process (which might be an oversight by MS?)

Well it's kernel cet and not user space cet, but who knows maybe they'll add support for it in some next insider builds and I did all this for nothing 😢

dmex:
@TheEragon 

> It won't look as clean and tidy

How are these changes? https://github.com/processhacker/processhacker/commit/aa0c066e4c2dd7cbc2967c6f93620b12a796e843

We already have to `== SYSTEM_PROCESS_ID` in other cases so this should be more consistent in that aspect?

> maybe they'll add support for it in some next insider builds and I did all this for nothing 😢

We still support previous versions of Windows so we would still be using these methods when that happens.

TheEragon:
Looks nice, I'll test it in a few hours.

PR #914, opened 2021-06-09
Fix a thread handle leak in the Process window Threads tab

jgottula:
Pretty sure this is a thread handle leak. The other code paths (e.g. for priority) call `NtClose`; but this particular one (critical) doesn't.


dmex (approved):

dmex:
> Pretty sure this is a thread handle leak

Yep it's a leak. Good catch!

The original code does close the handle but that line wasn't copied over for some reason when it was merged:
https://github.com/processhacker/plugins-extra/blob/a313bee6d290c5c0b6bf2ae04d1d16b688eed366/SetCriticalPlugin/main.c#L159-L171

PR #925, opened 2021-06-29
Update phnt

diversenok:
I adjusted the comments for token info classes that were missing the corresponding types, plus marked settable ones.


dmex (approved):

dmex:
LGTM 

Thanks 👍

Biswa96:
Why does phnt have its own PHNT_VERSION? Can't version constant from sdkddkver.h be used there?

dmex:
> Can't version constant from sdkddkver.h be used there

No. sdkddkver doesn't get maintained or updated properly sometimes and other times there are hundreds of releases available before the version name gets added and then MS doesn't set it properly.

It's easier, more consistent and more reliable using our own version define that we can change and update without Microsoft approval.

PR #974, opened 2021-08-29
Compilation fix

ge0rdi:
This will fix KPH compilation.
And also PH compilation for ARM64.

### DIFF
--- a/phlib/basesup.c
+++ b/phlib/basesup.c
@@ -612,7 +612,7 @@ SIZE_T PhCountStringZ(
     _In_ PWSTR String
     )
 {
-#if (PH_LEGACY_CRT_SUPPORT && !_ARM64_)
+#if (PH_LEGACY_CRT_SUPPORT && !defined(_ARM64_))
     if (PhpVectorLevel >= PH_VECTOR_LEVEL_SSE2)
     {
         PWSTR p;
@@ -1433,7 +1433,7 @@ ULONG_PTR PhFindCharInStringRef(
 
     if (!IgnoreCase)
     {
-#if (PH_LEGACY_CRT_SUPPORT && !_ARM64_)
+#if (PH_LEGACY_CRT_SUPPORT && !defined(_ARM64_))
         if (PhpVectorLevel >= PH_VECTOR_LEVEL_SSE2)
         {
             SIZE_T length16;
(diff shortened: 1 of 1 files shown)


phnt/include/ntbcd.h
```diff
@@ -21,6 +21,8 @@
 #ifndef _NTBCD_H
 #define _NTBCD_H
 
+#include <minwindef.h>
```
dmex, line 24: Why does this need to be imported?

phnt/include/ntexapi.h
```diff
@@ -22,6 +22,7 @@
 #define _NTEXAPI_H
 
 #include <ntkeapi.h>
+#include <ntimage.h>
```
dmex, line 25: Why does this also need to be imported?

phnt/include/ntbcd.h
ge0rdi: Because of `CALLBACK` definition.

phnt/include/ntexapi.h
ge0rdi: Because of `PIMAGE_EXPORT_DIRECTORY` used in `SYSTEM_GDI_DRIVER_INFORMATION` structure.

ge0rdi: Hmm, now I noticed that `ntimage.h` is only part of WDK.
And PHNT's `ntexapi.h` is used also in ProcessHacker (non-WDK) project.
So this change will make PH not compilable without WDK.

Will that be a problem?
Or WDK is required either way?

ge0rdi: Eventually we can use forward declaration instead of `ntheader.h` include:
`typedef struct _IMAGE_EXPORT_DIRECTORY * PIMAGE_EXPORT_DIRECTORY;`

What do you think?

phnt/include/ntbcd.h
dmex: I've deleted the callback type.

phnt/include/ntexapi.h
dmex: We don't use that structure so I changed the type

dmex:
I've fixed the other issues. The only change required is the one for `!defined(_ARM64_))`

ge0rdi:
> I've fixed the other issues. The only change required is the one for !defined(_ARM64_))

Thank you.

Updated PR.
Now it just fixes ARM64 build.

dmex (approved):

dmex:
> Now it just fixes ARM64 build.

Merged.

Btw.. These !_ARM64_ checks did work but regressed after we enabled the standards conforming compiler so I guess defined() is the only reliable check.

PR #985, opened 2021-09-02
Fix ARMv7 process detection

Biswa96:


ProcessHacker/proctree.c
```diff
@@ -3276,7 +3276,7 @@ BOOLEAN NTAPI PhpProcessTreeNewCallback(
                             PhInitializeStringRef(&getCellText->Text, L"x64");
                             break;
                         }
-                    case IMAGE_FILE_MACHINE_ARM:
+                    case IMAGE_FILE_MACHINE_ARMNT:
```
jxy-s, line 3279: probably just add `IMAGE_FILE_MACHINE_ARMNT`:

```cpp
case IMAGE_FILE_MACHINE_ARM:
case IMAGE_FILE_MACHINE_ARMNT:
```

Biswa96: Would not that be incorrect? According to the [docs][1],

> All code executed on this platform is expected to start and always remain in Thumb mode


[1]: https://docs.microsoft.com/en-us/cpp/build/overview-of-arm-abi-conventions

jxy-s: Yeah you're right, this is fine. Thank for fixing it.

Issue #938, opened 2021-07-28
[Feature Request] Middle-clicking a process should select all its "peers"


Yaron10:
**Is your feature request related to a problem? Please describe.**
I often use the great new feature [ToolStatus: Add working set for selected processes](https://github.com/processhacker/processhacker/commit/5a7672788811daebb35516ce81a021d6fdca6567) (thanks again! 👍) for summing up the total WS of a program running multiple processes.

I need to manually select all those processes.

**Describe the solution you'd like**
Middle-clicking a process should select all its "peers". 
That is: middle-clicking `firefox.exe` should select all processes with the same name.

Thank you.

Yaron10:
@dmex,

May I ask what you think of this suggestion?
I just want to know if I should be looking forward to the implementation.

Thank you.

Yaron10:
@TheEragon,

Could you please have a look at this request? - I'd appreciate it.

TheEragon:
I can certainly look at this if I find some time, but @dmex should first approve such change. he is the owner of this project.

Yaron10:
@TheEragon,
Thank you. Much appreciated.

@dmex,
I assumed you were busy and this feature should be very useful for me (and probably many other users).
Could you approve and assign it to @TheEragon?

Thank you for your great work.

dmex:
> middle-clicking firefox.exe should select all processes with the same name.
>  @dmex should first approve such change

@TheEragon Why my approval? 😂I might be the current maintainer but never used to be. You or anyone else can become the maintainer just by showing initiative and being self assertive. You don't need permission, act like the boss, you want something then make the changes. If you're consistent enough then you'll be given an official role like any other project 👍

> I assumed you were busy and this feature should be very useful for me (and probably many other users).

I personally feel using the existing feature when holding shift/control to select multiple items is what you should be using? The majority of people on Windows are accustomed to holding down shift/control when selecting multiple items, using mouse3 wouldn't be getting used very often so the changes might not be useful? I'm not sure so just been waiting for more comments/feedback.

TheEragon:
> @TheEragon Why my approval? 😂I might be the current maintainer but never used to be. You or anyone else can become the maintainer just by showing initiative and being self assertive. You don't need permission, act like the boss, you want something then make the changes. If you're consistent enough then you'll be given an official role like any other project 👍

Because I didn't feel like making a major GUI design change is my call :)

[...]

PR #1066, opened 2021-12-21
Add support for selecting process's subtree by middle clicking it

TheEragon:
This should fix the feature request #938

### DIFF
--- a/ProcessHacker/proctree.c
+++ b/ProcessHacker/proctree.c
@@ -4068,6 +4068,30 @@ BOOLEAN NTAPI PhpProcessTreeNewCallback(
             }
         }
         return FALSE;
+
+    case TreeNewMiddleClick:
+        // selecting subtree works only in NoSortOrder
+        if (ProcessTreeListSortOrder != NoSortOrder)
+            break;
+
+        PPH_TREENEW_MOUSE_EVENT mouseEvent = Parameter1;
+        if (!mouseEvent)
+            break;
+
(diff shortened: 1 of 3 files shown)


ProcessHacker/proctree.c
```diff
@@ -4068,6 +4068,30 @@ BOOLEAN NTAPI PhpProcessTreeNewCallback(
...
+        if (ProcessTreeListSortOrder != NoSortOrder)
+            break;
+
+        PPH_TREENEW_MOUSE_EVENT mouseEvent = Parameter1;
+        if (!mouseEvent)
+            break;
+
+        node = (PPH_PROCESS_NODE)mouseEvent->Node;
+
+        // init last index to self, this way we select only the process if there are no children
```
dmex, line 4083: please add `(TheEragon)` at the end of these comments so we know who wrote them :P

TheEragon: Will do :)

TheEragon:
I just realized that this doesn't work like one might except, right now it simply toggles subtrees that you middle click on. But it should probably deselect already selected items and select just current subtree. CTRL+middle click should add subtree to already selected processes, this way it will be consistent with left clicking, what do you think?

dmex:
> should probably deselect already selected items

Actions will apply to those items unless they're deselected.

dmex (approved):

dmex:
👍

PR #1067, opened 2021-12-21
Coherency fixes

ge0rdi:
There was an option for coherency computation introduced recently.
Some places weren't checking whether coherency is enabled and thus wrong information was shown in such case.


dmex:
Thanks, looks like I forgot to change a few places 😘

Issue #1069, opened 2021-12-22
Toolstatus's customization windows doesn't have proper theme support


TheEragon:
Toolstatus's customization window has a listbox without proper theme support

![image](https://user-images.githubusercontent.com/5801389/147088055-2a4282bd-3472-44ab-b49f-0ade7982405f.png)

I tried adding WM_CTLCOLORBTN, WM_CTLCOLORDLG and WM_CTLCOLORSTATIC but that doesn't help, maybe you'll know where the problem is, I'm not theme support expert 😅

OrbbQ3:
`WM_CTLCOLORLISTBOX` ?

TheEragon:
> `WM_CTLCOLORLISTBOX` ?

Thanks but no 😞, but I solved it. It's waiting for an approval.

PR #1073, opened 2021-12-22
ToolStatus theme updates

TheEragon:
- fixed a potential leak of `context->BrushNormal`
- simplified the code a bit
- implemented status bar's listbox theme support (#1069)
- fixed a flickering in status bar and tool bar, here if you moved the mouse rapidly around selected item in list box a flickering happened.


dmex:
> fixed a flickering in status bar and tool bar, here if you moved the mouse rapidly around selected item in list box a flickering happened.

This is caused by the WM_MOUSEMOVE redrawing in `PhpThemeWindowListBoxControlSubclassProc` for redrawing the listbox window border (light/dark) when the mouse enters and leaves the window. I was going to disable that code because the border highlight sometimes remains active when the mouse enters/leaves too quickly (also causing the flickering)?

dmex (approved):

PR #1076, opened 2021-12-27
Improve process heaps counters

ge0rdi:
`RtlQueryProcessDebugInformation` doesn't return information about
allocated bytes on segment heap.
Also it doesn't return proper values (allocated/committed) if big (virtual) allocations are
present on standard NT heap.

We are enumerating all heap entries already, so we can count these
statistics on our own.

This way we should show more accurate information for both NT and
segment heaps.

### DIFF
--- a/phlib/native.c
+++ b/phlib/native.c
@@ -10316,13 +10316,34 @@ NTSTATUS PhQueryProcessHeapInformation(
     {
         PRTL_HEAP_INFORMATION heapInfo = &debugBuffer->Heaps->Heaps[i];
         HANDLE processHandle;
+        SIZE_T allocated = 0;
+        SIZE_T committed = 0;
+
+        // go through all heap entries and compute amount of allocated and committed bytes
+        for (ULONG e = 0; e < heapInfo->NumberOfEntries; e++)
+        {
+            PRTL_HEAP_ENTRY entry = &heapInfo->Entries[e];
+
+            if (entry->Flags & RTL_HEAP_BUSY)
+                allocated += entry->Size;
+            if (entry->Flags & RTL_HEAP_SEGMENT)
+                committed += entry->u.s2.CommittedSize;
+        }
+
+        // sometimes computed number if commited bytes is few pages smaller than the one reported by API, lets use the higher value
+        if (committed < heapInfo->BytesCommitted)
+            committed = heapInfo->BytesCommitted;
+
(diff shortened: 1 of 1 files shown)


ge0rdi:
I have tested this only on Win10 (21H2), though I think it should work also on older systems.

phlib/native.c
```diff
@@ -10316,13 +10316,34 @@ NTSTATUS PhQueryProcessHeapInformation(
...
+        if (committed < heapInfo->BytesCommitted)
+            committed = heapInfo->BytesCommitted;
+
+        // make sure number of allocated bytes is not higher than number of committed bytes (as that would make no sense)
+        if (allocated > committed)
+            allocated = committed;
 
         heapDebugInfo->Heaps[i].Flags = heapInfo->Flags;
         heapDebugInfo->Heaps[i].Signature = ULONG_MAX;
         heapDebugInfo->Heaps[i].NumberOfEntries = heapInfo->NumberOfEntries;
```
ge0rdi, line 10343: I was thinking that this should be also adjusted to count only _busy_ entries.

The API returns various kinds of other heap entries (internal segments, free blocks, uncommitted ranges).
I think it would make much more sense to report just number actual _user_ allocations as that is probably what people may be usually interested in.

dmex:
>  it doesn't return proper values (allocated/committed) if big (virtual) allocations are present on standard NT heap.

I could be wrong but very large allocations via HeapAlloc bypass the heap entirely as a performance optimization.

> I think it would make much more sense to report just number actual user allocations as that is probably what people may be usually interested in.

It should report all types of blocks because they indicate different types of issues.

phlib/native.c
```diff
@@ -10316,13 +10316,34 @@ NTSTATUS PhQueryProcessHeapInformation(
...
+            PRTL_HEAP_ENTRY entry = &heapInfo->Entries[e];
+
+            if (entry->Flags & RTL_HEAP_BUSY)
+                allocated += entry->Size;
+            if (entry->Flags & RTL_HEAP_SEGMENT)
+                committed += entry->u.s2.CommittedSize;
+        }
+
+        // sometimes computed number if commited bytes is few pages smaller than the one reported by API, lets use the higher value
+        if (committed < heapInfo->BytesCommitted)
```
dmex, line 10334: You need to add (ge0rdi) to these comments.

ge0rdi: So every comment should contain reference to its author?
What is it good for?
I mean one can easily see author of given change in git's history.

PR #1077, opened 2021-12-28
Recognize all heap regions in process memory tab

### LINKED ISSUE #1050: PH does not display heap segments
The memory tab does not show that some private regions are used by the heap. (in the "Use" column).

The "Sysinternals VMMap" application shows that these memory regions are being used by the heaps.

The simplest example for diagnostics - memory region of the "notepad.exe" buffer. The location of this buffer should be on the heap, but if you find the entered string in memory, it will be in an unnamed private memory region. (**VMMap shows that this region is being used by the heap**).

In addition, PH does not display the "Portable Executable" sections and TEB like the VMMap does.

OS: Win11.

ge0rdi:
`RtlQueryProcessDebugInformation` can return information about each heap
memory segment. So we can use that to properly mark heap related memory regions.

This only works for native processes because `RtlQueryProcessDebugInformation`
doesn't work well for WoW64 processes.

Fixes #1050.

### DIFF
--- a/ProcessHacker/memprv.c
+++ b/ProcessHacker/memprv.c
@@ -244,58 +244,79 @@ void PhpUpdateHeapRegions(
     _In_ HANDLE ProcessHandle
     )
 {
-    PRTL_DEBUG_INFORMATION debugBuffer;
+    PRTL_DEBUG_INFORMATION debugBuffer = NULL;
     HANDLE powerRequestHandle = NULL;
 
     if (WindowsVersion >= WINDOWS_8)
     {
-        // The RtlQueryProcessDebugInformation can cause deadlock when querying frozen (suspended) immersive process.
-        // A workaround was implemented using PhCreateExecutionRequiredRequest().
+        // RtlQueryProcessDebugInformation can cause deadlock on some older versions when querying frozen (suspended) immersive process. (ge0rdi)
+        // A workaround was implemented using PhCreateExecutionRequiredRequest(). (ge0rdi)
         PhCreateExecutionRequiredRequest(ProcessHandle, &powerRequestHandle);
     }
 
-    if (debugBuffer = RtlCreateQueryDebugBuffer(0x10000, FALSE))
+    for (ULONG i = 0x10000; ; i *= 2) // based on PhQueryProcessHeapInformation (ge0rdi)
     {
-        if (NT_SUCCESS(RtlQueryProcessDebugInformation(List->ProcessId, RTL_QUERY_PROCESS_HEAP_SUMMARY | RTL_QUERY_PROCESS_HEAP_ENTRIES_EX, debugBuffer)))
+        if (!(debugBuffer = RtlCreateQueryDebugBuffer(i, FALSE)))
+            break;
+
+        NTSTATUS status = RtlQueryProcessDebugInformation(
+            List->ProcessId,
+            RTL_QUERY_PROCESS_HEAP_SUMMARY | RTL_QUERY_PROCESS_HEAP_ENTRIES_EX,
+            debugBuffer
+        );
+
+        if (!NT_SUCCESS(status))
         {
-            if (debugBuffer->Heaps)
+            RtlDestroyQueryDebugBuffer(debugBuffer);
+            debugBuffer = NULL;
+        }
+
+        if (NT_SUCCESS(status) || status != STATUS_NO_MEMORY)
+            break;
+
+        if (2 * i <= i)
+            break;
+    }
+
+    if (powerRequestHandle)
+        PhDestroyExecutionRequiredRequest(powerRequestHandle);
+
+    if (!debugBuffer)
+        return;
+
+    if (debugBuffer->Heaps)
+    {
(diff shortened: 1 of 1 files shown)


ProcessHacker/memprv.c
```diff
@@ -239,6 +239,65 @@ PPH_MEMORY_ITEM PhpSetMemoryRegionType(
...
+    if (WindowsVersion >= WINDOWS_8)
+    {
+        // The RtlQueryProcessDebugInformation can cause deadlock when querying frozen (suspended) immersive process.
+        // A workaround was implemented using PhCreateExecutionRequiredRequest().
+        PhCreateExecutionRequiredRequest(ProcessHandle, &powerRequestHandle);
+    }
+
+    if (debugBuffer = RtlCreateQueryDebugBuffer(0x10000, FALSE))
+    {
+        if (NT_SUCCESS(RtlQueryProcessDebugInformation(List->ProcessId, RTL_QUERY_PROCESS_HEAP_SUMMARY | RTL_QUERY_PROCESS_HEAP_ENTRIES_EX, debugBuffer)))
```
ge0rdi, line 259: `RTL_QUERY_PROCESS_HEAP_ENTRIES_EX` flag (0x200) does in fact return just information about heap segments (instead of all busy/free allocations).
Thus maybe `RTL_QUERY_PROCESS_HEAP_SEGMENTS` would be more appropriate name?

ProcessHacker/memprv.c
```diff
@@ -239,6 +239,65 @@ PPH_MEMORY_ITEM PhpSetMemoryRegionType(
...
+    _In_ PPH_MEMORY_ITEM_LIST List,
+    _In_ HANDLE ProcessHandle
+    )
+{
+    PRTL_DEBUG_INFORMATION debugBuffer;
+    HANDLE powerRequestHandle = NULL;
+
+    if (WindowsVersion >= WINDOWS_8)
+    {
+        // The RtlQueryProcessDebugInformation can cause deadlock when querying frozen (suspended) immersive process.
```
ge0rdi, line 252: This comment/workaround is based on comment in `PhQueryProcessHeapInformation`.

I was wondering if there is more info about the issue?
Like what versions are affected and if the problem is still present in recent Win10/Win11.

ge0rdi:
Here is quick comparison with my test application:

Before:
heap-before
After:
heap-after

ProcessHacker/memprv.c
dmex: Don't hard-code the commit size with 0x10000... It's required to loop and increment just like kernel32 does with Heap32First/Heap32Next: 

https://github.com/processhacker/processhacker/blob/b228cb0d2aa8035ae57a765c555bb1cf70be1de5/phlib/native.c#L10267-L10270

> HEAP_SEGMENTS would be more appropriate name

Yeah we can rename it.

ProcessHacker/memprv.c
dmex: > is more info about the issue?
> what versions are affected and if the problem is still present in recent Win10/Win11.

RtlQueryProcessDebugInformation creates a remote thread for the process but when that process was UWP/Immersive and frozen (suspended) the process/thread cannot be resumed and would deadlock.... It was only an issue on earlier versions of Windows 10 and was fixed by Microsoft.

The function was supposed to create an execution required power request and unfreeze the immersive process before creating the remote thread. So even though Microsoft fixed that issue, some users don't update Windows, so we still have to create our own power request before calling the function.

dmex (changes requested):
* Add (ge0rdi) after any comments.
* Loop the commit size.

ProcessHacker/memprv.c
ge0rdi: > Don't hard-code the commit size with 0x10000... It's required to loop and increment just like kernel32 does with Heap32First/Heap32Next:

Actually I wanted to make the code simpler here. Since we are interested only in heap segments, the returned debug buffer is usually quite small even for processes that use heap heavily (it contains just few records for each heap).

So it will practically always fit the 64k reserved debug buffer (which also seems to be smallest value that works with `RtlQueryProcessDebugInformation`).

Of course I can copy the code from `PhQueryProcessHeapInformation` if you think it would be better.

Though in that case it may be better to extract it to some helper function so that it can be use in both `PhQueryProcessHeapInformation` and `PhpUpdateHeapRegions`.
Not sure what would be the best place for such _helper_.

ge0rdi: @dmex 
> Yeah we can rename it.

Will you do that?
Or should I do it (within this PR)?

dmex: > So it will practically always fit the 64k reserved debug buffer

I had to debug this function and flags last week because of insider build changes and fixed commit sizes just doesn't work with some processes. The size loop was directly copied from Heap32First/Heap32Next in kernel32.dll, so the correct usage of this function requires the loop.

> Or should I do it (within this PR)?

 I'll update the flags separately because there are other changes required for insider builds:

image

dmex (approved):

PR #1085, opened 2022-01-04
Fix heap info processing on Win11 preview

ge0rdi:
Recent Win11 preview builds introduced new member into RTL_HEAP_INFORMATION structure.
Thus we will need to process heap info array differently based on OS build.


dmex:
I've been testing a patch which does this.

ge0rdi:
> I've been testing a patch which does this.

Cool. Looking forward to it 👍

Issue #1105, opened 2022-01-20
NetworkTools build failure


skipik:
### Brief description of your issue

![123](https://user-images.githubusercontent.com/24621645/150364049-ba6df003-7b6a-40ac-bbda-8d6243aa89a7.png)

PR #1106, opened 2022-01-20
Fix NetworkTools compilation after commit fc02d38b8

TheEragon:
#1105


skipik:
That's what we have now (after the fix in this PR): [ERROR] (1) D:\Dev\ProcessHacker3\src\trunk\plugins\NetworkTools\country.c(613,24): error C2198: 'PhImageListDrawIcon': too few arguments for call [D:\Dev\ProcessHacker3\src\trunk\plugins\NetworkTools\NetworkTools.vcxproj]

PR #1107, opened 2022-01-21
Theme: Enable dark theme for HANDLE properties subpages

TheEragon:
Here is how it looks like now

Event
New
image
Old
image

And security page
New
image
Old
image

The check marks look a bit weird in my opinion maybe with a bit of hacking we could make the background around it transparent but then it would be hard to read against this dark background 🤷‍♂️, what do you think?

### DIFF
--- a/ProcessHacker/include/phapp.h
+++ b/ProcessHacker/include/phapp.h
@@ -874,6 +874,12 @@ VOID PhShowTokenProperties(
     _In_opt_ PWSTR Title
     );
 
+INT CALLBACK PhpTokenSheetProc(
+    _In_ HWND hwndDlg,
+    _In_ UINT uMsg,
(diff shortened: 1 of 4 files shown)


dmex:
>  we could make the background around it transparent but then it would be hard to read against this dark background

The plan was to make these fixed property dialogs resizable but I couldn't figure out how to make the builtin security tab resize even with subclassing... So I was going to eventually replace the security tab with a custom tab that does resize (and would be easier to theme)?

TheEragon:
> The plan was to make these fixed property dialogs resizable but I couldn't figure out how to make the builtin security tab resize even with subclassing... So I was going to eventually replace the security tab with a custom tab that does resize (and would be easier to theme)?
> 

Great, I think that will be better than hacking I'm thinking about.

PR #1113, opened 2022-01-28
Fix compiler errors for ARM64 build

jdpatdiscord:
My understanding is this code loads the image then reads the exception entries. Unfortunately in this case, the only runtime function entry struct that `winnt.h` loads is for the current architecture being compiled for. This pull request brings both AMD64 and ARM64 runtime function entry definitions into Process Hacker's headers in order to fix errors due to struct entries that are missing.


dmex:
👍

PR #1132, opened 2022-02-20
Update KPH for VS2022

dmex:
Updates KPH for VS2022

The KPH version number was never updated after previous commits so this commit bumps the version to 3.2 but it should probably be changed to something higher?

Also updates the code style to match the [HACKING.md](https://github.com/processhacker/processhacker/blob/master/HACKING.md) conventions.

### DIFF
--- a/KProcessHacker/KProcessHacker.vcxproj
+++ b/KProcessHacker/KProcessHacker.vcxproj
@@ -43,6 +43,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <TargetVersion>Windows7</TargetVersion>
@@ -51,6 +52,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <TargetVersion>Windows7</TargetVersion>
@@ -59,6 +61,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'" Label="Configuration">
     <TargetVersion>Windows10</TargetVersion>
@@ -67,6 +70,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <TargetVersion>Windows7</TargetVersion>
@@ -75,6 +79,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|ARM64'" Label="Configuration">
     <TargetVersion>Windows10</TargetVersion>
@@ -83,6 +88,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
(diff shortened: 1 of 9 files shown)


KProcessHacker/main.c
```diff
@@ -395,30 +395,33 @@ NTSTATUS KpiPopulateKnownDllExtents(
...
+
+    status = ZwOpenSection(
+        &sectionHandle,
+        SECTION_MAP_READ | SECTION_QUERY,
+        &objectAttributes
+        );
+
     if (!NT_SUCCESS(status))
-    {
-        sectionHandle = NULL;
```
jxy-s, line 409: Please don't trust that garbage won't make it into the output parameter a failed call, while normally this isn't a big deal and in most circumstances the Windows API won't populate it with garbage on a failed call, it's good practice for kernel code to clear the output data that might land in cleanup later. The reasoning behind this pattern is to help ensure robustness of the kernel code.

KProcessHacker/process.c
```diff
@@ -266,17 +263,12 @@ NTSTATUS KpiOpenProcessToken(
         );
 
     if (!NT_SUCCESS(status))
-    {
-        process = NULL;
```
jxy-s, line 270: same here, I expect many of these got removed - please put them back

KProcessHacker/process.c
```diff
@@ -266,17 +263,12 @@ NTSTATUS KpiOpenProcessToken(
...
-        goto Exit;
-    }
-
-    primaryToken = PsReferencePrimaryToken(process);
+        goto CleanupExit;
 
-    if (!primaryToken)
+    if (!(primaryToken = PsReferencePrimaryToken(process)))
```
jxy-s, line 268: I really try to avoid assignment in conditionals. It's a risky pattern, there are even are compiler options to disallow them, and my analyzers will mark them as "code smell". That said, if this is a pattern we want to allow - fine to leave this change in.

KProcessHacker/verify.c
```diff
@@ -136,25 +130,37 @@ NTSTATUS KphHashFile(
...
+        NULL,
+        FILE_ATTRIBUTE_NORMAL,
+        FILE_SHARE_READ,
+        FILE_OPEN,
+        FILE_NON_DIRECTORY_FILE | FILE_SYNCHRONOUS_IO_NONALERT,
+        NULL,
+        0
+        )))
     {
         goto CleanupExit;
```
jxy-s, line 154: Not part of this diff, but please null the output `fileHandle` on failure.

KProcessHacker/verify.c
```diff
@@ -642,51 +663,56 @@ NTSTATUS KpiVerifyProcess(
...
     }
-    sectionHeaders = ExAllocatePoolWithTag(PagedPool,
-                                           sizeOfSectionHeaders,
-                                           'pvpK');
-    if (!sectionHeaders)
+
+    if (!(sectionHeaders = ExAllocatePoolZero(PagedPool, sizeOfSectionHeaders, 'pvpK')))
     {
-        goto Exit;
+        status = STATUS_ACCESS_DENIED;
```
jxy-s, line 678: This should be consistent with the above `ExAllocatePoolZero` ([line 398](https://github.com/processhacker/processhacker/pull/1132/files#diff-6c9dc9ccbf75e62ca944a08a1582237d348db77324191bc00eec3fc9bdc06c8eR398)). My preference would be `STATUS_INSUFFICIENT_RESOURCES` since this is an allocation failure and it's best to communicate that as such. There are other areas in this file that return `STATUS_INSUFFICIENT_RESOURCES` too.

KProcessHacker/dynimp.c
```diff
@@ -37,11 +37,8 @@ VOID KphDynamicImport(
 {
     PAGED_CODE();
 
-    KphDynPsGetProcessProtection = (PPS_GET_PROCESS_PROTECTION)(
-                       KphGetSystemRoutineAddress(L"PsGetProcessProtection"));
-
-    KphDynRtlImageNtHeaderEx = (PRTL_IMAGE_NT_HEADER_EX)(
-                           KphGetSystemRoutineAddress(L"RtlImageNtHeaderEx"));
+    KphDynPsGetProcessProtection = (PPS_GET_PROCESS_PROTECTION)KphGetSystemRoutineAddress(L"PsGetProcessProtection");
```
jxy-s, line 40: In general I try to keep under 80 character width, no more that 100, for the sake of viewing the code. Formatting changes feel odd to go into a commit titled "Update KPH for VS2022".

KProcessHacker/KProcessHacker.vcxproj
```diff
@@ -51,6 +52,7 @@
     <PlatformToolset>WindowsKernelModeDriver10.0</PlatformToolset>
     <ConfigurationType>Driver</ConfigurationType>
     <DriverType>WDM</DriverType>
+    <Driver_SpectreMitigation>Spectre</Driver_SpectreMitigation>
```
jxy-s, line 55: 👍 - I already have this in a the private branch for the hardening, happy to see this get in now.

KProcessHacker/verify.c
```diff
@@ -113,16 +113,10 @@ NTSTATUS KphHashFile(
 
     if (!NT_SUCCESS(status = BCryptOpenAlgorithmProvider(&hashAlgHandle, KPH_HASH_ALGORITHM, NULL, 0)))
         goto CleanupExit;
-    if (!NT_SUCCESS(status = BCryptGetProperty(hashAlgHandle, BCRYPT_OBJECT_LENGTH,
-        (PUCHAR)&hashObjectSize, sizeof(ULONG), &querySize, 0)))
-    {
+    if (!NT_SUCCESS(status = BCryptGetProperty(hashAlgHandle, BCRYPT_OBJECT_LENGTH, (PUCHAR)&hashObjectSize, sizeof(ULONG), &querySize, 0)))
```
jxy-s, line 116: Probably best to break these parameters onto their own line IMO.

KProcessHacker/verify.c
```diff
@@ -377,48 +390,50 @@ NTSTATUS KpiVerifyProcessSections(
...
     offset.QuadPart = textSection->PointerToRawData;
-    while (remainingBytes > 0)
+
+    while (remainingBytes != 0)
     {
         bytesToRead = FILE_BUFFER_SIZE;
-        if (remainingBytes < bytesToRead)
-        {
+
+        if (bytesToRead > remainingBytes)
```
jxy-s, line 409: Trying to keep me on my toes? 😛 - reversing this make it harder to read IMO

KProcessHacker/verify.c
```diff
@@ -377,48 +390,50 @@ NTSTATUS KpiVerifyProcessSections(
...
         status = STATUS_INSUFFICIENT_RESOURCES;
-        goto Exit;
+        goto CleanupExit;
     }
 
     remainingBytes = inspectBytes;
     offset.QuadPart = textSection->PointerToRawData;
-    while (remainingBytes > 0)
+
+    while (remainingBytes != 0)
```
jxy-s, line 405: diff didn't pick up on `>` changing to `!=` - probably due to the new line, making it harder to review

jxy-s (changes requested):
Please address the output parameter item(s). My hardening branch already went through and cleaned up a bunch of those.

The other things are more or less nits. I would prefer we try to eliminate the assignment in conditionals, at least from the kernel code - too easy to create bugs that way IMO.

jxy-s (approved):

PR #1138, opened 2022-02-23
Updater: Change color of string text to a readable one in dark theme

TheEragon:
I think this looks better, but feel free to change it to something else 😄 

image

### DIFF
--- a/plugins/Updater/options.c
+++ b/plugins/Updater/options.c
@@ -621,7 +621,10 @@ INT_PTR CALLBACK TextDlgProc(
                                         if (PhEqualString(commitHash, shortCommitHash, TRUE))
                                         {
                                             newFont = context->ListViewBoldFont;
-                                            customDraw->clrText = RGB(0, 0, 0x0);
+                                            if (PhGetIntegerSetting(L"EnableThemeSupport"))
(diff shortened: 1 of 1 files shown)


Biswa96:
Just wondering. Is is possible to use GetSysColor? There are many hand written color values.

plugins/Updater/options.c
```diff
@@ -621,7 +621,10 @@ INT_PTR CALLBACK TextDlgProc(
                                         if (PhEqualString(commitHash, shortCommitHash, TRUE))
                                         {
                                             newFont = context->ListViewBoldFont;
-                                            customDraw->clrText = RGB(0, 0, 0x0);
+                                            if (!PhGetIntegerSetting(L"EnableThemeSupport"))
```
dmex, line 624: `If not` doesn't make sense. Use this instead:

```
if (PhGetIntegerSetting(L"EnableThemeSupport"))
    customDraw->clrText = RGB(125, 125, 125);
else
    customDraw->clrText = GetSysColor(COLOR_WINDOWTEXT);                           
                                                
```

dmex (changes requested):

dmex (approved):

dmex:
👍

PR #1168, opened 2022-03-25
Update used LDRP flags in modlist.c & modprv.c.

evelynmarie:
Fixes build issues when attempting to build the latest commit of Process Hacker & the plugins, as these files were still referencing the older `LDRP_IMAGE_NOT_AT_BASE` constant rather than `LDRP_COR_DEFERRED_VALIDATE`.


dmex:
> LDRP_COR_DEFERRED_VALIDATE 

This flag isn't the same as LDRP_IMAGE_NOT_AT_BASE (which is deprecated)

> Fixes build issues when attempting to build the latest commit 

All of this code is getting removed because it's incorrect on Windows 7 and aboce. I'm testing a few things before committing these changes 👍

PR #1180, opened 2022-04-13
Update KPH solution

dmex:
Add latest solution properties
Fix KpiOpenThreadProcess regression on Windows 7 and 8
Add workaround for Microsoft blocking the driver from testsigning mode on Windows 11


KProcessHacker/thread.c
```diff
@@ -283,13 +268,29 @@ NTSTATUS KpiOpenThreadProcess(
...
+    
+    // Note: Windows 7 and Windows 8 require KernelMode (dmex)
     status = ObOpenObjectByPointer(
         PsGetThreadProcess(thread),
         0,
         NULL,
         DesiredAccess,
         *PsProcessType,
-        AccessMode,
+        KernelMode,
```
jxy-s, line 293: A note for observers. This is an intentional decision to bypass access checks allowing for inspection of the thread. This aligns this path with the existing code: https://github.com/processhacker/processhacker/blob/master/KProcessHacker/thread.c#L164

The call is restricted by the key validation and stricter yet for write access requiring domination and caller validation.

jxy-s (approved):

PR #1221, opened 2022-05-16
Recognize more well-known memory regions

diversenok:
There are a few pointers in PEB that we can use to find and mark some well-known memory regions on the memory tab.


jxy-s:
Changes look fine to me. Only observation I have is the behavior is changing slightly so we don't so information if it is "null". I think that is reasonable though.

jxy-s (approved):
Changes look fine to me. Only observation I have is the behavior is changing slightly so we don't so information if it is "null". I think that is reasonable though.

PR #1265, opened 2022-06-16
Update SECURITY.md

jxy-s:


stelladtheresome:
I don't understand all this but I'm being attacked by a hacker and I think my dogs are too. I could use some help

stelladtheresome:
I don't understand all of this but I'm being attacked by a hacker and I
think my dogs are to.. Darla VanGinkel

Issue #1278, opened 2022-06-17
Bug: Standard Context Menu lag when launched as admin (Menu bar or Tray icon)


AltF5:
### Brief description of your issue

Video demo of bug:
**https://www.youtube.com/watch?v=NkShxsJwrQA**

### Steps to reproduce

I'll attach this Debug64 build (unzip, rename .txt to .exe, verify im not trying to pwn you ... which I wouldn't based on how long i've been around here).
[SystemInformer.zip](https://github.com/winsiderss/systeminformer/files/8929020/SystemInformer.zip)

> Run elevated
> Click on the '**System**' menubar item (not View | Tools | Users | or Help which are Unaffected)
> Also impacted: Tray icon right clicking.

Does the lag occur to you with this build, or is it truly all my Win10 Machines & VMs when launched elevated ? (I don't have a Win11 VM at the moment to confirm). I don't quite understand how this would not be reproducible for you. This is occuring with the code downloaded as-is compiled in Debug | x64 as-is, with no settings, with no modifications, and attempted both in VS2022 and VS2022 Preview (both which compile fine, but experience this Menu bug).
**Can you try backing up and deleting your settings in %AppData%\Process Hacker just to be sure you don't have some setting that is preventing this?**

### Actual behavior
https://www.youtube.com/watch?v=NkShxsJwrQA

### Expected behavior
Function as intended. Not like that ^

### Environment

Win10 21H2.
Latest Windows SDK from VS2022 (did not ask for an upgrade, so nothing was changed)

dmex:
I can't reproduce lag opening these menus... If anything your security software is screwing with bcd.dll and the registry BCD keys.

AltF5:
@dmex sorry but this makes no sense (at least to me :) ) regarding **bcd.dll** and **registry BCD keys**
1. What does BCD and bcd.cpp have to do with the context menu generation (unless it is involved?) ?
2. Windows Defender is entirely disabled from the driver & kernel perspective on one of the PCs.
3. This can be repro-ed on multiple PCs, VMs, etc when running as an admin regardless of "security software

I'll attach this build (unzip, rename .txt to .exe, verify im not trying to pwn you ... which I wouldn't based on how long i've been around here)
[SystemInformer.zip](https://github.com/winsiderss/systeminformer/files/8928931/SystemInformer.zip)
"

jxy-s:
[...]

PR #1282, opened 2022-06-17
Init once firmware and boot lists

jxy-s:
This is to minimize the cost of showing the boot and firmware entries in the shutdown/reboot options. Fix for: https://github.com/winsiderss/systeminformer/issues/1278


AltF5:
FYI -- unfortunately this does not correct the delay issue with the 'System' menu (or Tray icon). Not sure if more is still yet to be done, but I've replaced actions.c with this and commented back in the 2 problematic menus, and the delay still occurs

jxy-s:
@AltF5 - just spoke with with @dmex on this and caching the entries isn't an appropriate solution anyway.

PR #1290, opened 2022-06-19
Add monospace font setting & use in memory list

rtldg:
image
~~left aligned vs right aligned~~

~~I would also love to make the base address column use a monospace font (something like `GetStockObject(SYSTEM_FIXED_FONT)`) but I couldn't find a good place to squeeze that in~~

Updated pr info:
Monospace font option added. It's used by the `Base Address` column in memory and modules tabs. The right-alignment is likely to be removed soon & zero-padding to be added, as suggested by jxy-s.


jxy-s:
It might be best to keep it left-justified and zero-fill in the formatting. I would also appreciate the monospace font change.

jxy-s:
Please rebase the branch to master. I will be closing the PR if no progress is made.

rtldg:

jxy-s:
Thanks @rtldg for updating the PR. I think this overall the change does make it easier to read. @dmex and I believe it might not make sense since it's the first column. I think ultimately we should do mono-space font, zero-pad the values, and left align. Did you end up finding a good place to squeeze in your suggestion (`GetStockObject(SYSTEM_FIXED_FONT)`)?

rtldg:
> Did you end up finding a good place to squeeze in your suggestion (`GetStockObject(SYSTEM_FIXED_FONT)`)?

Yeah, sorry I got caught up with something else before I could make a commit to show what I did after doing the CLA thing.

I didn't want to start adding parameters to functions & there didn't seem to be a good way to check what a node's column was, so I ended up adding a new #define bitflag that I passed to `PhAddTreeNewColumn` to be used in `PhTnpDrawCell`.

jxy-s:
I don't completely follow `TextFlags` are system flags. We are already up to `Ex2` on that API. We could hijack the high bits of `Alignment` as flags for "mono-space text alignment" and mask off the high bits for the existing paths. Hate to have to add an `Ex3` 😉.

rtldg:
> I don't completely follow `TextFlags` are system flags.

I used the high bit (0x80000000) and later strip it out of the `textFlags` variable passed to `DrawText`. https://github.com/winsiderss/systeminformer/commit/89eb57b49218402f282889ccdf5809f63dbd7214
It's not the cleanest solution of course, but it worked without too much API muck.
POC commit I guess since I'm sure it might not be preferred to do it this way.

rtldg:
Sample pic of it on Windows 10 right now (with the right alignment included & no zero-padding).
image

jxy-s:
Looks like the `DrawText` flags haven't changed in a while, it might be okay for now to hijack them. It's safest to put the new flags on `Alignment`, looks like we treat it as as flags anyway.

```cpp
    if (Column->Alignment & PH_ALIGN_LEFT)
        item.fmt |= HDF_LEFT;
    else if (Column->Alignment & PH_ALIGN_RIGHT)
        item.fmt |= HDF_RIGHT;
    else
        item.fmt |= HDF_CENTER;
```

Also it might be best to specifically try to choose a better mono-space font. `SYSTEM_FIXED_FONT` looks a bit aggressive compared to the rest of the fonts there.

rtldg:
Moved the flag to `Alignment`. Definitely a better idea.
Swapped out `GetStockObject(SYSTEM_FIXED_FONT)` with `CreateFontA(...FF_MODERN...)` which will use some random font that is a bit better.
The font size (for whatever random font this picked) is still not quite matching with the rest of the rows, but it's a prettier font for sure. I'm not quite sure which variables to poke to make that better.
Could do some CSS-esque font enumerations or searches for specific, "better" fonts & fallbacks if you want. And `modlist.c` could also have the `Base address` column use a monospace font, but https://github.com/winsiderss/systeminformer/pull/1290/commits/f64f9fa16e8dd35c0e29ce2b2c1fca7d5eeae37e for now for feedback.
image

phlib/include/treenew.h
```diff
@@ -10,6 +10,8 @@ extern "C" {
 #define PH_TREENEW_SEARCH_TIMEOUT 1000
 #define PH_TREENEW_SEARCH_MAXIMUM_LENGTH 1023
 
+#define PH_TREENEW_FIXED_WIDTH_FONT 0x80000000
```
jxy-s, line 13: I suggest moving this next to the other `PH_ALIGN_xxx` defines in `guisup.h` name it with the same prefix, suggesting `PH_ALIGN_MONOSPACE_FONT`.

phlib/treenew.c
```diff
@@ -5576,10 +5576,36 @@ VOID PhTnpDrawCell(
...
+    static HFONT fixedWidthFont = NULL;
 
     font = Node->Font;
     textFlags = Column->TextFlags;
 
+    if (Column->Alignment & PH_TREENEW_FIXED_WIDTH_FONT)
+    {
+        if (fixedWidthFont == NULL)
+        {
+            fixedWidthFont = CreateFontA(
```
jxy-s, line 5588: With the caveat that I am _not_ a UI guy 😄: 

Take a look at `PhDuplicateFont` and `PhDuplicateFontWithNewWeight`. It might be possible to duplicate the current font and specify `FF_MODERN`? Or maybe just use `PhCreateFont`L"Consolas", ...);` with the same weight and height as the current font?

jxy-s: Maybe also a problem of double-assign of the `static` here? Or perhaps the UI thread here is guaranteed to be serialized already (my lack of UI experience is showing).

rtldg: > Maybe also a problem of double-assign of the `static` here? Or perhaps the UI thread here is guaranteed to be serialized already (my lack of UI experience is showing).

I did write it with the assumption that it'd either be single-threaded for UI sections or that a duplicate font handle would be a negligible memory leak. Of course it was even less negligible previously when it was `GetStockObject(SYSTEM_FIXED_WIDTH)` which presumably uses some sort of object interning or static handles since `GetStockObject` says `"It is not necessary (but it is not harmful) to delete stock objects by calling DeleteObject."`

Also, I'll make a commit for the review comments sometime later today/currenttime+12h or something

jxy-s: Thanks for going the extra mile here. We'll chat more later.

It may also make sense to make the "monospace" font a global that gets initialized like the other fonts the application uses. Then possibly add the ability for users to choose/configure the font like the others are. This will make it easier to drop monospace into other areas too.

SystemInformer/main.c
```diff
@@ -564,6 +564,25 @@ VOID PhInitializeFont(
     if (oldFont) DeleteFont(oldFont);
 }
 
+VOID PhInitializeMonospaceFont(
+    VOID
+)
```
jxy-s, line 569:
```suggestion
    )
```

SystemInformer/mainwnd.c
```diff
@@ -3633,6 +3635,36 @@ VOID PhMwpInvokeUpdateWindowFont(
     if (oldFont) DeleteFont(oldFont);
 }
 
+VOID PhMwpInvokeUpdateWindowFontMonospace(
+    _In_opt_ PVOID Parameter
+)
```
jxy-s, line 3640:
```suggestion
    )
```

SystemInformer/options.c
```diff
@@ -714,6 +716,27 @@ static BOOLEAN GetCurrentFont(
     return result;
 }
 
+_Success_(return)
+static BOOLEAN GetCurrentFontMonospace(
+    _Out_ PLOGFONT Font
+)
```
jxy-s, line 722:
```suggestion
    )
```

jxy-s (commented):
Just some style suggestions from me. This looks reasonable and the result, to me, is easier to read with the monospace font.

I think there is still the outstanding question of if we want to left-justify and zero-pad. But that could (and maybe should) be done in a follow-up PR.

I'm going to hold off on my approval until @dmex has had an opportunity to review, he's the expert in this area.

rtldg:
> I think there is still the outstanding question of if we want to left-justify and zero-pad. But that could (and maybe should) be done in a follow-up PR.

Some options for that that wouldn't require adding new functions:
```diff
             case PHMMTLC_BASEADDRESS:
-                PhPrintPointer(node->BaseAddressText, memoryItem->BaseAddress);
+#ifdef _WIN64
+                swprintf(node->BaseAddressText, sizeof(node->BaseAddressText) / sizeof(WCHAR), L"%016llx", (UINT64)memoryItem->BaseAddress);
+#else
+                swprintf(node->BaseAddressText, sizeof(node->BaseAddressText) / sizeof(WCHAR), L"%08x", (UINT32)memoryItem->BaseAddress);
+#endif
```
Or something like this
```diff
             case PHMMTLC_BASEADDRESS:
-                PhPrintPointer(node->BaseAddressText, memoryItem->BaseAddress);
+                for (int i = 0; i < sizeof(int*) * 2; i++)
+                {
+                    node->BaseAddressText[i] = L"0123456789ABCDEF"[0xF & ((INT_PTR)memoryItem->BaseAddress >> ((sizeof(int*)*2-i-1) * 4))];
+                }
```

> I'm going to hold off on my approval until dmex has had an opportunity to review, he's the expert in this area.

👍

jxy-s:
I think it's most appropriate to put in a new function in `phsup.h` alongside `PhPrintPointer` for printing a pointer with zero-padding. Your lookup table solution is likely to be faster than `printf`. I recommend using lower case alpha characters so it's consistent with the exiting. And don't forget the null terminator.

I think we could also make a general improvement to `PhPrintPointer`. The integer conversion functions are general purpose. The bit shifting and table lookup, from my measurements, is faster (nanoseconds):

```
PhPrintPointer (old)    PhPrintPointer (new)    PhPrintPointerEx
300                           100                             0
400                           100                           100
500                           100                             0
600                           100                             0
200                             0                             0
500                           100                           100
500                             0                             0
200                           100                             0
400                             0                             0
300                             0                             0
400                           100                             0
400                           100                           100
500                             0                             0
600                           100                             0
600                             0                             0
300                           100                           100
300                           100                             0
400                           100                             0
500                           100                           100
400                           100                             0
400                           100                             0
400                           100                           100
400                           100                             0
500                             0                             0
500                           100                           100
```

```cpp
#define PH_PRINT_ZERO_PAD 0x00000001ul

VOID PhPrintPointerEx(
    _Out_writes_(PH_PTR_STR_LEN_1) PWSTR Destination,
    _In_ PVOID Pointer,
    _In_ ULONG Flags
    )
{
    PUCHAR bytes;
    USHORT j;
    BOOLEAN start;

    j = 0;

    start = ((Flags & PH_PRINT_ZERO_PAD) != 0);

    Destination[j++] = L'0';
    Destination[j++] = L'x';

    bytes = (PUCHAR)&Pointer;
    bytes = &bytes[sizeof(PVOID) - 1];

    for (USHORT i = 0; i < (sizeof(PVOID) * 2); i += 2, bytes--)
    {
        if (!start && ((*bytes >> 4) != 0))
        {
            start = TRUE;
        }

        if (start)
        {
            Destination[j++] = L"0123456789abcdef"[*bytes >> 4];
        }

        if (!start && ((*bytes & 0xf) != 0))
        {
            start = TRUE;
        }

        if (start)
        {
            Destination[j++] = L"0123456789abcdef"[*bytes & 0xf];
        }
    }

    Destination[j++] = L'\0';
}

VOID PhPrintPointer(
    _Out_writes_(PH_PTR_STR_LEN_1) PWSTR Destination,
    _In_ PVOID Pointer
    )
{
    PhPrintPointerEx(Destination, Pointer, 0);
}
```

jxy-s:
This has been sitting for a while. @dmex suggested to me to look into `PhFormatToBuffer` which should eliminate us having to write a new format function for this.

rtldg:
> This has been sitting for a while. 

Yeah, sorry about that

> dmex suggested to me to look into `PhFormatToBuffer` which should eliminate us having to write a new format function for this.

Made a commit using that for zero-padding the hexadecimal pointer.
Added `PhInitFormatIXPadZeros` in it, which is `PhInitFormatIX` with the `FormatPadZeros` flag included.

There were merge conflicts (that didn't actually exist?) that didn't allow the merge-by-rebase button on github to work, so I used a merge commit for updating the branch.

SystemInformer/modlist.c
```diff
@@ -86,7 +86,7 @@ VOID PhInitializeModuleList(
 
     // Default columns
     PhAddTreeNewColumn(Context->TreeNewHandle, PHMOTLC_NAME, TRUE, L"Name", 100, PH_ALIGN_LEFT, -2, 0);
-    PhAddTreeNewColumn(Context->TreeNewHandle, PHMOTLC_BASEADDRESS, TRUE, L"Base address", 80, PH_ALIGN_RIGHT, 0, DT_RIGHT);
+    PhAddTreeNewColumn(Context->TreeNewHandle, PHMOTLC_BASEADDRESS, TRUE, L"Base address", 80, PH_ALIGN_RIGHT | PH_ALIGN_MONOSPACE_FONT, 0, DT_RIGHT);
```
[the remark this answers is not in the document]
jxy-s: There is also a few other locations in `modlist.c` and `modprv.c` that needs updated to use the new formatting:

https://github.com/winsiderss/systeminformer/blob/master/SystemInformer/modprv.c#L647
https://github.com/winsiderss/systeminformer/blob/master/SystemInformer/modlist.c#L1034-L1047

https://github.com/winsiderss/systeminformer/blob/master/SystemInformer/modlist.c#L108-L109

I thought about just changing `PhPrintPointer` but I'm hesitant since it's used in some other places too. Let's just focus on these locations since we're touching the `Modules` and `Memory` tabs of the process information.

jxy-s (changes requested):

SystemInformer/modlist.c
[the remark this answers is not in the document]
rtldg: Yeah, it'd be definitely be convenient if `PhPrintPointer` did handle it since I feel a bit bad about copy-pasting the same code around.

SystemInformer/modprv.c
```diff
@@ -515,13 +515,18 @@ VOID PhModuleProviderUpdate(
...
+    SIZE_T returnLength;
 
     // If we didn't get a handle when we created the provider,
     // abort (unless this is the System process - in that case
     // we don't need a handle).
     if (!moduleProvider->ProcessHandle && moduleProvider->ProcessId != SYSTEM_PROCESS_ID)
         goto UpdateExit;
 
+    PhInitFormatS(&format[0], L"0x");
+    PhInitFormatIXPadZeros(&format[1], 0);
```
jxy-s, line 528: Move this down to where it is used, please - then you can remove the assignment to `UIntPtr`.

jxy-s (changes requested):

SystemInformer/modprv.c
rtldg: Are you saying to move only the `PhInitFormatIXPadZeros` or both format functions (& the variables `format`/`returnLength`)?

jxy-s: Move both down before `PhFormatToBuffer`. This way you don't have to look up to hunt for the formatting that is going to be applied to the `PhFormatToBuffer(.., 2..`.

jxy-s (approved):
LGTM

I pulled the branch and did some testing. The alignment isn't perfect with the default monospace font. With the font mixing that is expected. There might be a better default monospace we could choose that better fits with the default font.

I played around with some of the `DT_` alignment options for the monospace font and also tried some others. In some cases dropping the size from 9 to 8 helps. And some other monospace fonts might be better. But I think the one you've chosen for the default is reasonable/passable for now. I'm willing to accept these changes and make refinements as necessary.

Thanks for going to extra mile on this!

PR #1315, opened 2022-07-23
Fix process private/shared commit statistics

ge0rdi:
Currently it shows the information only for processes in a job.
And actually it is not correct information at all.
Because it contains grouped commit statistics for all processes in the job.

We will use `VM_COUNTERS_EX2` to obtain correct information for process on modern systems.
And use job info as fall-back for older systems.

### DIFF
--- a/phlib/include/phnativeinl.h
+++ b/phlib/include/phnativeinl.h
@@ -878,24 +878,51 @@ FORCEINLINE
 NTSTATUS
 PhGetProcessAppMemoryInformation(
     _In_ HANDLE ProcessHandle,
-    _Out_ PPROCESS_JOB_MEMORY_INFO JobMemoryInfo
+    _Out_ PAPP_MEMORY_INFORMATION AppMemoryInfo
     )
 {
     NTSTATUS status;
-    PROCESS_JOB_MEMORY_INFO jobMemoryInfo;
+    APP_MEMORY_INFORMATION appMemoryInfo = {0};
+    VM_COUNTERS_EX2 vmCounters;
 
-    // Win32 called this ProcessAppMemoryInfo with APP_MEMORY_INFORMATION struct (dmex)
     status = NtQueryInformationProcess(
         ProcessHandle,
-        ProcessJobMemoryInformation,
-        &jobMemoryInfo,
-        sizeof(PROCESS_JOB_MEMORY_INFO),
+        ProcessVmCounters,
+        &vmCounters,
+        sizeof(vmCounters),
         NULL
-        );
+    );
 
     if (NT_SUCCESS(status))
     {
-        *JobMemoryInfo = jobMemoryInfo;
+        appMemoryInfo.PrivateCommitUsage = vmCounters.CountersEx.PagefileUsage;
+        appMemoryInfo.TotalCommitUsage = appMemoryInfo.PrivateCommitUsage + vmCounters.SharedCommitUsage;
+        appMemoryInfo.PeakPrivateCommitUsage = vmCounters.CountersEx.PeakPagefileUsage;
+    }
+    else
+    {
+        // fall-back to Job memory info on older systems
+        PROCESS_JOB_MEMORY_INFO jobMemoryInfo;
+
+        status = NtQueryInformationProcess(
+            ProcessHandle,
+            ProcessJobMemoryInformation,
+            &jobMemoryInfo,
+            sizeof(jobMemoryInfo),
+            NULL
+        );
+
+        if (NT_SUCCESS(status))
+        {
+            appMemoryInfo.PeakPrivateCommitUsage = jobMemoryInfo.PeakPrivateCommitUsage;
+            appMemoryInfo.PrivateCommitUsage = jobMemoryInfo.PrivateCommitUsage;
+            appMemoryInfo.TotalCommitUsage = jobMemoryInfo.PrivateCommitUsage + jobMemoryInfo.SharedCommitUsage;
+        }
+    }
+
+    if (NT_SUCCESS(status))
+    {
+        *AppMemoryInfo = appMemoryInfo;
     }
 
     return status;
--- a/phlib/include/phnativeinl.h
+++ b/phlib/include/phnativeinl.h
@@ -874,15 +874,23 @@ PhSetProcessBreakOnTermination(
         );
 }
 
+typedef struct _PH_APP_MEMORY_INFORMATION
+{
+    ULONG64 AvailableCommit;
+    ULONG64 PrivateCommitUsage;
+    ULONG64 PeakPrivateCommitUsage;
+    ULONG64 TotalCommitUsage;
+} PH_APP_MEMORY_INFORMATION, * PPH_APP_MEMORY_INFORMATION;
+
 FORCEINLINE
 NTSTATUS
 PhGetProcessAppMemoryInformation(
     _In_ HANDLE ProcessHandle,
-    _Out_ PAPP_MEMORY_INFORMATION AppMemoryInfo
+    _Out_ PPH_APP_MEMORY_INFORMATION AppMemoryInfo
     )
 {
     NTSTATUS status;
-    APP_MEMORY_INFORMATION appMemoryInfo = {0};
+    PH_APP_MEMORY_INFORMATION appMemoryInfo = {0};
     VM_COUNTERS_EX2 vmCounters;
 
     status = NtQueryInformationProcess(
--- a/SystemInformer/prpgstat.c
+++ b/SystemInformer/prpgstat.c
@@ -311,7 +311,7 @@ VOID PhpUpdateProcessStatistics(
         if (Context->ProcessHandle)
         {
             PH_PROCESS_WS_COUNTERS wsCounters;
-            APP_MEMORY_INFORMATION appMemoryInfo;
+            PH_APP_MEMORY_INFORMATION appMemoryInfo;
 
             if (NT_SUCCESS(PhGetProcessWsCounters(Context->ProcessHandle, &wsCounters)))
             {
--- a/SystemInformer/prpgstat.c
+++ b/SystemInformer/prpgstat.c
@@ -311,7 +311,7 @@ VOID PhpUpdateProcessStatistics(
         if (Context->ProcessHandle)
         {
             PH_PROCESS_WS_COUNTERS wsCounters;
-            PROCESS_JOB_MEMORY_INFO appMemoryInfo;
+            APP_MEMORY_INFORMATION appMemoryInfo;
 
             if (NT_SUCCESS(PhGetProcessWsCounters(Context->ProcessHandle, &wsCounters)))
             {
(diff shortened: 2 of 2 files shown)


ge0rdi:

phlib/include/phnativeinl.h
```diff
@@ -820,24 +820,51 @@ FORCEINLINE
...
+
+    if (NT_SUCCESS(status))
+    {
+        appMemoryInfo.PrivateCommitUsage = vmCounters.CountersEx.PagefileUsage;
+        appMemoryInfo.TotalCommitUsage = appMemoryInfo.PrivateCommitUsage + vmCounters.SharedCommitUsage;
+        appMemoryInfo.PeakPrivateCommitUsage = vmCounters.CountersEx.PeakPagefileUsage;
+    }
+    else
+    {
+        // fall-back to Job memory info on older systems
```
jxy-s, line 846: Why use the query failure as an indication that the info query is not supported on a platform? If we know where is began support we should check `WindowsVersion` and make the appropriate call.

It is also unclear to me if we should even keep querying `ProcessJobMemoryInformation` if it's inaccurate.

ge0rdi: > If we know where is began support we should check WindowsVersion and make the appropriate call.

I'm not sure when support for `VM_COUNTERS_EX2` was added exactly.
Do you know?

> Should we even keep querying ProcessJobMemoryInformation if it's inaccurate?

Yup, that's right. I'd personally remove it.
I just assumed you guys would like to keep at least some info on older systems (if it was like that for long time).

I'd gladly remove the fallback. Just let me know if it should stay or not.

diversenok: > I'm not sure when support for VM_COUNTERS_EX2 was added exactly.
Do you know?

Just checked, it was added between Windows 8.1 and Windows 10 TH1.

ge0rdi: > Just checked, it was added between Windows 8.1 and Windows 10 TH1.

Thanks.
I guess there should be no need for exact OS version check.
We can check for `STATUS_INFO_LENGTH_MISMATCH` that the API returns when new (unsupported) structure is supplied.

@jxy-s 
Do we want to keep the (improper) fallback? Or rather just remove it?

jxy-s: I'm okay with removing it, it seems errant to me. But I'd like input from @dmex since he's the author.

phlib/include/phnativeinl.h
```diff
@@ -820,24 +820,51 @@ FORCEINLINE
...
 {
     NTSTATUS status;
-    PROCESS_JOB_MEMORY_INFO jobMemoryInfo;
+    APP_MEMORY_INFORMATION appMemoryInfo = {0};
+    VM_COUNTERS_EX2 vmCounters;
 
-    // Win32 called this ProcessAppMemoryInfo with APP_MEMORY_INFORMATION struct (dmex)
     status = NtQueryInformationProcess(
         ProcessHandle,
-        ProcessJobMemoryInformation,
```
dmex, line 832: GetProcessInformation with the `ProcessAppMemoryInfo` infoclass returns [APP_MEMORY_INFORMATION](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-app_memory_information) by calling NtQueryInformationProcess with `ProcessJobMemoryInformation`

phlib/include/phnativeinl.h
```diff
@@ -820,24 +820,51 @@ FORCEINLINE
 NTSTATUS
 PhGetProcessAppMemoryInformation(
     _In_ HANDLE ProcessHandle,
-    _Out_ PPROCESS_JOB_MEMORY_INFO JobMemoryInfo
+    _Out_ PAPP_MEMORY_INFORMATION AppMemoryInfo
```
dmex, line 823: replace PAPP_MEMORY_INFORMATION with a custom struct type:

```c
typedef struct _PH_APP_MEMORY_INFORMATION
{
    ULONG64 AvailableCommit;
    ULONG64 PrivateCommitUsage;
    ULONG64 PeakPrivateCommitUsage;
    ULONG64 TotalCommitUsage;
} PH_APP_MEMORY_INFORMATION, *PPH_APP_MEMORY_INFORMATION;
```

phlib/include/phnativeinl.h
ge0rdi: Yup, I know that.
And it is wrong. Because it returns wrong information (cumulated stats for all processes in the job).

On newer Win10 it uses `ProcessVmCounters` as fallback for processes that are not in job.
So at least for such processes the API returns correct information.

I think we don't need to stick to the Win API behavior, do we?

phlib/include/phnativeinl.h
ge0rdi: Done.

phlib/include/phnativeinl.h
jxy-s: Thanks @dmex, I understand the intention behind this now.

`GetProcessInformation` + `ProcessAppMemoryInfo` does a bit more than just query `NtQueryInformationProcess` with `ProcessJobMemoryInformation`.

I did some RE of this path and I think we want this. @ge0rdi could you try this out please?:

```cpp
NTSTATUS PhGetProcessAppMemoryInformation(
    _In_ HANDLE ProcessHandle,
    _Out_ PAPP_MEMORY_INFORMATION AppMemoryInfo
    )
{
    NTSTATUS status;
    ULONG64 availableCommit;
    APP_MEMORY_INFORMATION memoryInfo;
    PROCESS_JOB_MEMORY_INFO jobMemoryInfo;

    status = NtQueryInformationProcess(
        ProcessHandle,
        ProcessJobMemoryInformation, 
        &jobMemoryInfo,
        sizeof(jobMemoryInfo),
        NULL
        );

    if (status == STATUS_ERROR_PROCESS_NOT_IN_JOB)
    {
        VM_COUNTERS_EX2 vmCounters;

        status = NtQueryInformationProcess(
            ProcessHandle, 
            ProcessVmCounters, 
            &vmCounters,
            sizeof(vmCounters), 
            NULL
            );

        if (!NT_SUCCESS(status))
            return status;

        memoryInfo.PeakPrivateCommitUsage = 0;
        memoryInfo.PrivateCommitUsage = vmCounters.CountersEx.PagefileUsage;
        memoryInfo.TotalCommitUsage = vmCounters.SharedCommitUsage + vmCounters.CountersEx.PagefileUsage;
        availableCommit = 0;
    }
    else
    {
        memoryInfo.PeakPrivateCommitUsage = jobMemoryInfo.PeakPrivateCommitUsage;
        memoryInfo.PrivateCommitUsage = jobMemoryInfo.PrivateCommitUsage;
        memoryInfo.TotalCommitUsage = jobMemoryInfo.PrivateCommitUsage + jobMemoryInfo.SharedCommitUsage;
        if (jobMemoryInfo.TotalCommitLimit)
        {
            availableCommit = jobMemoryInfo.TotalCommitLimit
                - (jobMemoryInfo.PrivateCommitUsage
                   + jobMemoryInfo.SharedCommitUsage);
        }
        else if (jobMemoryInfo.PrivateCommitLimit)
        {
            availableCommit = jobMemoryInfo.PrivateCommitLimit - jobMemoryInfo.PrivateCommitUsage;
        }
        else
        {
            availableCommit = 0;
        }
    }

    if (!availableCommit)
    {
        SYSTEM_BASIC_PERFORMANCE_INFORMATION systemPerfInfo;

        status = NtQuerySystemInformation(
            SystemBasicPerformanceInformation,
            &systemPerfInfo,
            sizeof(systemPerfInfo),
            NULL
            );

        if (!NT_SUCCESS(status))
            return status;

        // There is a global in KernelBase which seems to be used as if it were PAGE_SIZE (jxy-s)
        availableCommit = PAGE_SIZE * (systemPerfInfo.CommitLimit - systemPerfInfo.CommittedPages);
    }

    memoryInfo.AvailableCommit = availableCommit;

    RtlCopyMemory(AppMemoryInfo, &memoryInfo, sizeof(memoryInfo));

    status = STATUS_SUCCESS;

    return status;
}
```

jxy-s: Or, keeping with what @dmex had originally, something that outputs `PROCESS_JOB_MEMORY_INFO` and not `APP_MEMORY_INFORMATION`.

jxy-s: Actually, taking a step back from this. I think the information @dmex is gathering is just fine. However, we should do a better job at displaying this information in the statistics. This isn't necessarily gathering application memory information. It is gathering the job information. And that is valuable information to show in a clear way in the statistics panel separate from the memory section, IMHO.

ge0rdi: So what about renaming current counters to `Job shared commit`, `Job private commit`, `Job peak private commit` and display them only if process is in job.
For obtaining those stats we can have `PhGetProcessJobMemoryInformation` that would fill `PROCESS_JOB_MEMORY_INFO` structure.

Then we can have `Shared commit` counter that will show process' shared commit on newer OSes (where it is supported).
This can be obtained by (new) `PhGetProcessVmCountersEx2` that will fill `VM_COUNTERS_EX2` structure.

I can prepare all this.
I'd just appreciate some guidance about naming, etc.

Eventually, if you guys want to take care of this, I can just fill issue regarding wrong Shared Commit statistics.

jxy-s: We should add a new section to the statistics section for "Job" which houses information like this, and may show "N/A" when the process is not in a job.

> For obtaining those stats we can have PhGetProcessJobMemoryInformation that would fill PROCESS_JOB_MEMORY_INFO structure.

Naming sounds good to me.

> This can be obtained by (new) PhGetProcessVmCountersEx2 that will fill VM_COUNTERS_EX2 structure.

It's might be time to update the `PH_PROCESS_ITEM` to have `VM_COUNTERS_EX2`. But I'd recommend doing that in a separate PR/commit. Or let us take care of this one as it's got some wide-reaching implications. I'm unsure if it generates much value outside of what we have today. All that to say it need some more time/thought.

With respect to what I'd like you to do for this PR. I would be nice to extend the statistics to have that job information and fill it with the `PROCESS_JOB_MEMORY_INFO` like I mentioned above. If you're willing to do that work I would appreciate it. Otherwise, yes, please open an issue on this topic and we'll take care of it when we have time.

ge0rdi: I've tried to add new section to process statistics:
image

Though honestly, it doesn't feel right :(
The statistics page is about statistics related to given process.
But here we are showing statistics related to job (different object representing bunch of processes in general).

More appropriate place would be on `Job` page:
image

But there are few issues with Job page:
- It is accessible only for administrators (because driver is needed to open job object)
- Current driver cannot open jobs with restricted ACLs, so that many processes that are in job don't even show Job page (your driver work PR should fix that afaik)

So I really don't know what to do :(

Well, we can start with simple `Job` section in process statistics (even if that's probably not the best way).
And then eventually do it more properly in the future.

I can update this PR with the change if you think it is the way to go.

jxy-s: Yeah I see your point. This stuff does really belong on the job statistics. I think it might be fine to put in on both 🤔. The advantage here is we could show some statistics related to the job the process is associated with without having to have the driver active. But in that regard we might want to just make the job tab more visible to show information that we can without the driver anyway.

I think in the end the job tab might need a little more love/attention. It likely hasn't received any since it's been broken due to the driver for a long time. I did check and my driver work will resolve the job problem. And after that time we can give the job tab some more attention.

I think it might be wisest to just hold off on theses changes for a little bit while we finish up the migration and driver work. Then revisit it once we have a better place to put this information in. We should at very least put an issue in to go back and revisit this once we're ready (if we don't have one already).

ge0rdi: > But in that regard we might want to just make the job tab more visible to show information that we can without the driver anyway.

Yup, that's exactly what I was thinking of.

> We should at very least put an issue in to go back and revisit this once we're ready (if we don't have one already).

I have created issues #1325 and #1326 for this.

I guess we can close this PR now.

PR #1320, opened 2022-07-26
driver work

jxy-s:
PR implements stricter requirements for making calls to the driver, hardens existing code, fixes SAL, contains mitigations against exploitation of the verified process, and expands the capabilities of the driver both around informing of system activity and exposing new functionality/information through the driver API.

The core of the changes here is the process state (defined in `kphapi.h`). Which we track in the driver and require a certain level of state for the connected client to have in order to be authorized to make a call into the driver (defined in `comms_handlers.c`).


KSystemInformer/KSystemInformer.vcxproj
```diff
@@ -33,7 +33,7 @@
     <Configuration>Debug</Configuration>
     <Platform Condition="'$(Platform)' == ''">Win32</Platform>
     <RootNamespace>KSystemInformer</RootNamespace>
-    <WindowsTargetPlatformVersion>$(LatestTargetPlatformVersion)</WindowsTargetPlatformVersion>
```
dmex, line 36: `$(LatestTargetPlatformVersion)`

dmex (changes requested):

KSystemInformer/KSystemInformer.vcxproj
jxy-s: Specific targeting is intentional for now. VS2022 and the newer SDKs don't play nicely with our targeted OSes and architectures.

ge0rdi:
@jxy-s 
Could you possibly add some _readme_ describing how to use compiled driver regarding SystemInformer custom signing?

I needed to debug some things that require driver (working with Jobs) and had to figure out how to do the signing properly.
I have used this quite outdated guide:
https://wj32.org/processhacker/forums/viewtopic.php?f=24&t=2852

Some up-to-date info would be really helpful.

jxy-s:
> @jxy-s Could you possibly add some _readme_ describing how to use compiled driver regarding SystemInformer custom signing?
> 
> I needed to debug some things that require driver (working with Jobs) and had to figure out how to do the signing properly. I have used this quite outdated guide: https://wj32.org/processhacker/forums/viewtopic.php?f=24&t=2852
> 
> Some up-to-date info would be really helpful.

Yes, happy to update documentation and migrate that info into the repo.

KSystemInformer/KSystemInformer.vcxproj
ge0rdi: What about at least using more recent WDK 22000 ?
That should still be able to target Windows7 without a problem.

FYI
VS2022 requires WDK 22621 (Win11 22H2) and that does support only Windows10 targets.
It also does support only x64 and arm64 builds.
https://docs.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk

I was wondering what are your plans regarding VS/WDK and `KSystemInformer`.
Do you want to stick with VS2019 and old WDK (like for long time)?
Or would you consider dropping Win7/8 (and x86) support?

jxy-s: > What about at least using more recent WDK 22000 ?

That's reasonable, I'll update to use `10.0.22000.0` with the next commit.

> VS2022 requires WDK 22621 

Yep, I'm very familiar with it. Driver development is usually lags behind the latest and greatest anyway. This is pretty standard and Microsoft historically screws up the WDK during transitions like this. I'm hopeful they'll walk back those restrictions a little bit, but we'll see.

> Do you want to stick with VS2019 and old WDK (like for long time)?
> Or would you consider dropping Win7/8 (and x86) support?

As long as we're going to support Win7/8 and x86 we'll stay on the WDK and build tools that Microsoft says supports it. Eventually we'll drop support for those legacy OSes. I don't see that happening immediately. For now I'm interested in getting 3.0 out and people migrated to it with minimal compatibility problems.

I have discussed alternatives with @dmex but I do not want to deal with the struggle of conditionally trying to support something that Microsoft explicitly states is not supported. We recently updated the other projects to have a dependency on the latest SDK and WDK. Which means we can't use `$(LatestTargetPlatformVersion)` for the driver. It is strictly, as of right now, incompatible with what we're targeting.

jxy-s:
> Yes, happy to update documentation and migrate that info into the repo.

This is done.

KSystemInformer/KSystemInformer.vcxproj
jxy-s: Updated target to `10.0.22000.0`.

KSystemInformer/README.md
```diff
@@ -0,0 +1,46 @@
...
+The driver tracks verified clients, restricts access by other actors on the 
+system, and denies access if the process is tampered with. The intention is to 
+discourage exploitation of the client when the driver is active. If tampering 
+or exploitation is detected the client is denied access.
+
+## Development
+
+Developers may generate their own key pair for use in their environment.
+
+1. execute `tools\CustomSignTool\bin\Release64\CustomSignTool.exe createkeypair public.key kph.key`
```
ge0rdi, line 34: The arguments should be `createkeypair kph.key public.key`.
Private key file should go first.

The old KPH guide had it wrong.

KSystemInformer/README.md
```diff
@@ -0,0 +1,46 @@
...
+2. copy `kph.key` into `tools\CustomSignTool\resources`
+3. copy the bytes for `public.key` into the `KphpTrustedPublicKey` array in [verify.c](verify.c)
```
ge0rdi, line 36: There is very nice _linux_ command that can format binary file into C array:
`xxd -i public.key`
It will print array that you can easily paste to `verify.c`.

It works with WSL (Windows Subsystem for Linux) right away.

KSystemInformer/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Once these steps are completed builds of System Informer components will 
+generate a `.sig` file next to the output binaries. And the developer built 
+driver will use the specified key when doing verification checks. Any plugins
+not built through the regular build process must also have their own `.sig`.
+
+Developers may suppress protections and state requirements by changing
```
ge0rdi, line 43: What about doing this for Debug driver builds automatically?
Do you see any issue with that?

jxy-s: I don't like having these types of "suppress protections" even exist to begin with. I want it to be a conscious decision to disable them. I personally prefer to actually test and develop using what is delivered 😄 - so I'd probably never turn this off and I'd opt to use a kernel debugger instead of a user mode debugger.

ge0rdi: > I don't like having these types of "suppress protections" even exist to begin with.

Are you afraid of some misuse?

I understand you want to have all protections in production driver as that will be used publicly.
But we are talking about private debug builds here.

> I want it to be a conscious decision to disable them.

Well, compiling Debug configuration pretty much means you are going to debug something, so you don't need any of these protections to get in a way.

> I'd opt to use a kernel debugger instead of a user mode debugger

Not everyone likes to torture himself :)
While I like WinDbg and use it a lot. I'd always prefer the comfort of VS user-mode debugging.

jxy-s: > compiling Debug configuration pretty much means you are going to debug something

And what if you need to debug the protections?

Seems like I'm the minority here. So I'll default it off for debug and comment it out myself rather than make others do it.

ge0rdi: > And what if you need to debug the protections?

You can explicitly define `KPH_PROTECTION_SUPPRESSED=0`.

> Seems like I'm the minority here. So I'll default it off for debug and comment it out myself rather than make others do it.

I didn't mean to press on you :)

Just wanted you to consider that maybe those protections could be disabled in Debug by default.
But I'm fine with any default, as long as it will be possible to disable it.

ge0rdi:
> ❌ @dmex

Hehe. It seems that @dmex needs to sign CLA too 😄

KSystemInformer/KSystemInformer.vcxproj
```diff
@@ -138,143 +138,179 @@
...
+      <WppTraceFunction>KphTracePrint(LEVEL,EVENT,MSG,...)</WppTraceFunction>
+      <WppFileExtensions>.c.cpp.h.hpp</WppFileExtensions>
+      <WppPreserveExtensions>.c.cpp.h.hpp</WppPreserveExtensions>
+      <DisableSpecificWarnings>4201</DisableSpecificWarnings>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>ksecdd.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>ksi.lib;FltMgr.lib;ksecdd.lib;kphlib_km.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <RandomizedBaseAddress>true</RandomizedBaseAddress>
       <AdditionalOptions>/INTEGRITYCHECK /BREPRO /DEPENDENTLOADFLAG:0x800 /PDBALTPATH:%_PDB% %(AdditionalOptions)</AdditionalOptions>
```
ge0rdi, line 222: I wanted to ask about `/PDBALTPATH:%_PDB%` option.
Is it really necessary to have it in Debug builds?

With the option there is only PDB name without full path inside driver binary. Which makes it harder to load symbols during debugging.
When full path is present, WinDbg loads symbols right away without any special action needed (otherwise I had to put it to concrete folder in my symbol cache).

Or do you know some other trick how to load symbols more easily?

KSystemInformer/README.md
jxy-s: Thinking about this a bit more. My concern here is that by forcing the protections off in debug mode it _will_ be wonky unless the developer deploys the debug mode driver _too_. I did put in some KPH warnings which pop up informing people _why_ access is being restricted, but that all breaks by doing the "default off in debug mode".

During this transition period to the new driver there will be some "getting used to the new workflow" regardless. I really think the right thing here is to not force it off in debug mode and force developers to actually _know_ what they're doing if they want to debug the project.

KSystemInformer/KSystemInformer.vcxproj
jxy-s: I usually just add the build output paths to `_NT_SYMBOL_PATH` so I never need to think about it again.

We could probably remove this truncation in debug builds and only have it in release builds. I'd have to look at the other `vcxproj` files too. Anyway, I think this ask it outside of the scope for this PR. If you like please open a ticket for it and we can discuss it there.

KSystemInformer/README.md
jxy-s: I'm going to resolve this conversation for now @ge0rdi - if it becomes an issue in people's workflow I'll happily change it in a later commit/PR.

KSystemInformer/README.md
jxy-s: There are lot of tools for this. I'm sure people are capable of converting the bits of the file to an array that can be pasted into verify.c.

KSystemInformer/KSystemInformer.vcxproj
jxy-s: I'm resolving discussion this for now. I'm happy to extend build support for VS2022 and the newer WDK when the time is appropriate.

A few parting notes:
1. VS2022 and the WDK extension is just broken right now there is a clearly posted regression on the front page of the WDK. Once Microsoft resolves that I will update the build script to provide an optional "latest" switch which chooses VS2022, forces the target to Windows10 and WDK, then omits Win32. This is similar to what @dmex had previously but due to the regression in VS2022 and the WDK extension I want to wait until that gets resolved.
2. Once we deprecate support for Win32 and Win 7/8 I will re-target and update the build script accordingly. Or, in the event Microsoft walks back WDK requirements before then, I will update the WDK version to use the latest.

KSystemInformer/KSystemInformer.vcxproj
ge0rdi: > I usually just add the build output paths to _NT_SYMBOL_PATH so I never need to think about it again.

Yup. But that requires manual action.

> Another alternative is to add some tooling to use symstore.exe to store output of build symbols automatically in a symbol store location of the developer's choosing.

That would be nice. But why to bother with that when all that is needed is to keep full PDB paths in binary.

> We could also integrate this into our build pipeline so we can have our own public symbol store 😄.

It would be really great if you can provide symbols for public builds (symbol server would be cool, but even some package that can be downloaded will be nice).

> We could probably remove this truncation in debug builds and only have it in release builds.

Other option would be to remove the linker option from projects completely and then _force_ the setting in build scripts by setting `LINK=/PDBALTPATH:%_PDB%` environment variable.
That way developers will have full local paths in binaries (which is fine) and official builds will have stripped paths.

> If you like please open a ticket for it and we can discuss it there.

Will do.
I just dealt with this recently and since I saw this driver overhaul PR I wanted to ask about it.

jxy-s:
I plan to merge this PR this weekend. Please attend to reviews if you're interested in doing so.

PR #1340, opened 2022-08-14
update modern icons

jxy-s:
This updates the modern icons and makes them show cleaner on the dark theme.

image

image


jxy-s:
"Always on top" is a bit wonky, I'll see if I can clean that up.

PR #1360, opened 2022-09-03
migrate poolmon to extendedtools

jxy-s:
This PR migrates PoolMon from the plugins-extra into ExtendedTools.

image

image

image


jxy-s:
poolmon needs updated for theme support

PR #1370, opened 2022-09-07
driver ps set apis

jxy-s:
Implements a limited number process executive set APIs through driver. These APIs are restricted to clients with maximum state and do not permit manipulation across a protected process domination boundary. This enables a client, when connected to the driver, to manipulate a small set of performance related properties about a thread or process when handle stripping may have otherwise prevented it.


yardenshafir (approved):

dmex (commented):
LGTM

PR #1372, opened 2022-09-08
Fix a typo

ruslangaripov:
According to the \`\`[PE Format](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#load-configuration-layout)'' article published on MSDN, the `GuardCFDispatchFunctionPointer` field of the load configuration structure is "the VA where Control Flow Guard **dispatch**-function pointer is stored".  Just like the `GuardXFGDispatchFunctionPointer` field below is a "**dispatch**-function".  I believe "*check*-function" was a typo in ldprp.c.


ruslangaripov:
I have read the CLA Document and I hereby sign the CLA.

jxy-s (approved):

PR #1387, opened 2022-09-24
Fix bug in network tab related to hash table.

MarekKnapek:
Hash table in network tab is configured to accept items of type PHP_RESOLVE_CACHE_ITEM in its constructor, but every other operation on it (except the constructor) is using element type of PPHP_RESOLVE_CACHE_ITEM instead. This causes stack overflow read when adding new item into it. The table is reading 32 bytes out of the stack, where only 8 bytes are prepared for it by its user.


MarekKnapek:

dmex:
Copypastas :(

Good catch 👍

PR #1394, opened 2022-09-28
Add -prerelease flag to vswhere invocations

CXuesong:
Fixes the issue that nothing happens and nothing shows in the console when calling `build_thirdparty.cmd` (and similar build scripts) when we only have VS Preview installed.

Also shows explicit message in the console if there is no VS installation detected.


CXuesong:

dmex (approved):

dmex:
LGTM 👍

PR #1439, opened 2022-10-21
Potential fix for compile time error indicating C4703, potentially un…

trparky:
…initialized local pointer variable 'deviceDescription'

### DIFF
--- a/plugins/HardwareDevices/gpuoptions.c
+++ b/plugins/HardwareDevices/gpuoptions.c
@@ -333,6 +333,7 @@ VOID FindGraphicsDevices(
         D3DKMT_HANDLE adapterHandle;
         PPH_STRING deviceDescription;
         PGPU_ENUM_ENTRY entry;
+        deviceDescription = NULL;
 
         entry = PhAllocateZero(sizeof(GPU_ENUM_ENTRY));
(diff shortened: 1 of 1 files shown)


trparky:
Including that one line of code, namely line number 336, seems to make it so that the plugin code will compile as versus not compiling at all.

The error that's produced without that line is...

[ERROR] (1) C:\Users\trpar\source\repos\systeminformer\plugins\HardwareDevices\gpuoptions.c(342): error C4703: potentially uninitialized local pointer variable 'deviceDescription' used [C:\Users\trpar\source\repos\systeminformer\plugins\HardwareDevices\HardwareDevices.vcxproj]LINK : fatal error LNK1257: code generation failed [C:\Users\trpar\source\repos\systeminformer\plugins\HardwareDevices\HardwareDevices.vcxproj]

trparky:

dmex:
> uninitialized local pointer variable 'deviceDescription'

This was a warnings generated by SAL annotations. Those functions initialize and allocate the string and not the callers was annotated with `_Success_(return)` but the value wasn't annotated with `_Out_opt_` even though it was optional so it was throwing a warning about the function returning an uninitialized value.

I've updated the annotations and fixed the warning 👍

PR #1474, opened 2022-11-09
Updated thirdparty lib binaries.

trparky:
Updated thirdparty lib binaries to reflect changes to the compiler in the last batch of updates deployed by Microsoft as part of the last Visual Studio update.


jxy-s:
Thanks for doing this. But I'm not comfortable checking in libs built by anyone but the primary maintainers. We will update them when we get an opportunity to.

PR #1477, opened 2022-11-10
Added clean commands to the build_thirdparty.cmd file.

trparky:
Added clean commands to the build_thirdparty.cmd file to fix situations where "fatal error C1047" might occur. Simply building the third-party lib files isn't enough, you have to do a clean operation first.


dmex:
I've added cleanup for build_thirdparty with commit a10b5cee7b3b20e1f4d909616c89dee3384665f7 👍

PR #1485, opened 2022-11-15
CustomBuildTool: Fixed build error if a "kph.key" file doesn't exist.

trparky:
I had error "[CopyKernelDriver] (1)" during compile, I determined that the cause of the error was because the "kph.key" file didn't exist. I included code to check for the existence of a "kph.key" file.

### DIFF
--- a/tools/CustomBuildTool/Build.cs
+++ b/tools/CustomBuildTool/Build.cs
@@ -171,7 +171,7 @@ namespace CustomBuildTool
                     }
                 }
             }
-            catch (Exception) {  }
+            catch (Exception) { }
 
             if (
                 string.IsNullOrWhiteSpace(BuildBranch) ||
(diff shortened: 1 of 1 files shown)


tools/CustomBuildTool/Build.cs
```diff
@@ -536,31 +536,36 @@ public static bool CopyKernelDriver(BuildFlags Flags)
...
+
+                if (File.Exists(PathToKPHKeyFile))
                 {
-                    if (File.Exists(file))
+                    foreach (var file in files)
                     {
-                        var sigfile = Path.ChangeExtension(file, ".sig");
-
-                        File.WriteAllText(sigfile, string.Empty);
+                        if (File.Exists(file))
```
jxy-s, line 545: I think the same check might need added in `SignPlugin` below.

dmex:
I made some changes and fixed in commit https://github.com/winsiderss/systeminformer/commit/f9c64c4f39548cd3225980544ea24c9d125e6bc2 (trparky as author) 👍

PR #1493, opened 2022-11-18
Few fixes

ge0rdi:
There is no point to provide "fake" handle to system API. Moreover it could cause crashes if "Strict handle check" mitigation is enabled.

### DIFF
--- a/phlib/symprv.c
+++ b/phlib/symprv.c
@@ -379,15 +379,18 @@ BOOL CALLBACK PhpSymbolCallbackFunction(
         {
             PIMAGEHLP_CBA_READ_MEMORY callbackData = (PIMAGEHLP_CBA_READ_MEMORY)CallbackData;
 
-            if (NT_SUCCESS(NtReadVirtualMemory(
-                ProcessHandle,
-                (PVOID)callbackData->addr,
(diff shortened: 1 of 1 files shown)


ge0rdi:
I'm running `SystemInformer` with `Application Verifier` enabled (to find some bugs eventually).
And opening `Memory` tab in `System information` causes immediate crash due to this bug (as AVRF enables `Strict handle check` mitigation).

ge0rdi:
Fixed also use-after-free bug caught by AVRF.

dmex:
Can you revert the change to PhpProcessHeapsDlgProc? I'm going to change the callback.

ge0rdi:
@dmex 
Done.

jxy-s (approved):

dmex:
> no point to provide "fake" handle to system API

Dbghelp.dll must have changed some of the checks for these callbacks... They're also passing uninitialized stack memory for the `FileName` value in the `CBA_DEFERRED_SYMBOL_LOAD_START` callback:

image


>  bug caught by AVRF

Fixed with 4ca0f90738a45239960ef19b4119205927290a4c

PR #1508, opened 2022-11-24
Display AppContainer integrity and not Low for AppContainers

TheEragon:
This is just an idea and feel free to close the PR if you don't like it, but it makes more sense to display integrity as `AppContainer` for stuff that runs inside container like Process Explorer does rather than `Low`.

image


dmex (approved):

dmex:
👍

PR #1509, opened 2022-11-25
ToolStatus: Fix process search cancelling

ge0rdi:
Properly deselect process nodes also if process search was cancelled (either by clicking 'x' button or providing empty string).

Otherwise it could happen that process that is part of collapsed tree will remain selected and that will cause various issues. For example properties for different process may be displayed, because SI thinks some other process is selected (than the one user clicked on).

### DIFF
--- a/plugins/ToolStatus/main.c
+++ b/plugins/ToolStatus/main.c
@@ -685,13 +685,10 @@ LRESULT CALLBACK MainWndSubclassProc(
                         // Cache the current search text for our callback.
                         PhSwapReference(&SearchboxText, newSearchboxText);
 
-                        if (!PhIsNullOrEmptyString(SearchboxText))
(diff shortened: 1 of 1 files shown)


ge0rdi:
Not sure if the description is understandable, but this is what I've experienced (and was furious at):

- I was looking for some process that is normally collapsed (I have `services` tree collapsed by default)
- Then clicked on the process to see its properties
- Once done with that I've cancelled the search (by clicking the `x` button), so that I have full tree again

And then I noticed that if I double click on some processes in the tree, properties for that previous process (I was searching for) were displayed instead of properties for the process that I've selected.

It was quite infuriating, because it seemed to be random.

dmex (approved):

dmex:
LGTM

PR #1512, opened 2022-11-25
peview: Fix window context use-after-free and memory leaks

ge0rdi:


ge0rdi:
In theory better fix would be to remove and free context in `WM_NCDESTROY` handling (apparently that comes after `WM_NOTIFY`).
But I think it should be enough to do it just in `WM_DESTROY` (yup, its simple and lazy :) ).

I can change it if you think it will be better that way.

dmex:
peview doesn't have memory leaks or use-after-free in these cases because the dialogs and windows are never destroyed - everything is left in memory until the application is terminated because the propsheet API doesn't destroy the windows.

FYI doing any memory freeing or cleanup in peview is completely pointless. I'll merge the changes but any memory freeing/cleanup will likely get removed in the future.

ge0rdi:
> peview doesn't have memory leaks or use-after-free in these cases because the dialogs and windows are never destroyed

Yup, leaks are not real problem (as peview is ending anyway).
I've just fixed them out of habit and also for consistency - some windows did cleanup of their context, some not.

Though the use-after-free issue was real.
When page window was destroyed, context was freed (but not removed from window property) and after that `WM_NOTIFY` was sent from child window and it accessed now destroyed context.
In most cases it doesn't cause any issue because the memory is most likely still there. But it could cause random crashes on peview shutdown.

> any memory freeing/cleanup will likely get removed in the future

If there will be no resource cleanup, then there can't be user-after-free.
So that sounds good as well :)

PR #1514, opened 2022-11-26
implemented process search filter marker to include all childen proce…

gmit3:
I've implemented a 'process list' 'search filter' marker that will include all children processes for a process matched by its name or pid.

For example, if you filter the process list with 'devenv', you'll end up with devenv.exe process(es) only. With my change, if you filter by 'devenv*', you'll end up with devenv.exe and all its children process subtree.

If * marker is not used, search will behave as before. Using * with multiple filters separated with | is fully supported.

As said in the beginning, * will be applied only on processes matched by process name or pid. Otherwise, a resulting tree would often end up with a mess.

If my change will be accepted, as a future enchantment, I would likely implement a + marker which would specify how many levels of children will be displayed. For example, 'devenv+' would display only immediate children, 'devenv++' would display two levels etc.


gmit3:

dmex:
1) The PR changes the filter callback type from BOOLEAN to an enum but only the one callback was updated to the enum while the others are still returning BOOLEAN/TRUE/FALSE?

2) Why were spaces removed from if statements?

image

gmit3:
1. No, that's not the case. The signature of a callback is still:
	BOOLEAN ProcessTreeFilterCallback

The original callback's code became a new function:
	static MatchType ProcessTreeFilterCallback2
as I needed to call it twice from the new code in the callback.

Having MatchType in all callbacks wouldn't make sense as it's specific for ProcessTree filter...


2. Yikes. I have autoformatting and my diff tool ignored whitespace changes. Sorry, I've committed the fix.

dmex (approved):

gmit3:
> I've merged the PR and made some changes to fix naming and improve performance: [ecc8906](https://github.com/winsiderss/systeminformer/commit/ecc8906e839c83d31c32fe58f68cfe14c9990f5f)
> 
> The wildcards are disabled by default and you need to set `EnableChildWildcardSearch` to enable. Let me know if those changes broke anything or if you think it's ready and stable and I'll change `EnableChildWildcardSearch` so wildcards are enabled by default 👍

Unfortunately, if the new option is enabled, filter doesn't seem to be working at all (3.0.6383), regardless if wildcard is specified or not.

dmex:
> if the new option is enabled, filter doesn't seem to be working

It was left enabled by default and setting 1 disabled the wildcard. I've updated the nightly to r6384 and fixed the setting.

PR #1529, opened 2022-12-10
Update native.c

DavidXanatos:
bugfix for PhGetProcessQuotaLimits


DavidXanatos:

DavidXanatos:
I have added an other fix as well:

The new Kernel driver returns the kernel stack trace as well as the user stack trace, 
without this fix System Informer shows the user stack duplicated when the driver is used.

dmex:
> The new Kernel driver returns the kernel stack trace as well as the user stack trace

@jxy-s was adding support for flags to enable/disable user stacks and they're not passed to the client yet:

https://github.com/winsiderss/systeminformer/blob/2fe3a6e71edc958e89ec7e3de96febdbb36eef1e/KSystemInformer/comms_handlers.c#L470-L473

DavidXanatos:
> 

I noticed that, but as its not yet done and at this place in the code we care only for the kernel part of the stack, I think this is a fair fix, and will not cause issues later.
If at some point the plan will be to get the entire stack from the driver only other things will need to be changed as well, at that point the fix can be simply removed.

jxy-s:
API change here: https://github.com/winsiderss/systeminformer/pull/1532 - what @DavidXanatos has done in the meantime is reasonable

PR #1596, opened 2023-02-08
Revert "Tidy up PhSetValueKeyZ"

trparky:
This reverts commit cd886e0a399e70212da71fe433fb276e4e172d97.

Something went wrong with this commit in which any setups that are created with this commit fail at this part...

and die.

Reverting this one single commit fixes things and allows for a setup to install properly.


dmex:
The setup wasn't calling the correct macro. Fixed with 0705c47624f70c2c5ffe93f82d7a58d2dc3939b3

PR #1649, opened 2023-03-18
Update ntexapi.h: NtDelayExecution sal fix

pokanoka:
NtDelayExecution `DelayInterval` argument is not actually optional, as can be seen in disasm in current kernels of ws2022, w11. I've also confirmed it being non-optional in disasm of win10 RTM.


dmex:
> I've also confirmed it being non-optional in disasm of win10 RTM

It was changed to optional when I reviewed some insider builds around 2018-2019 🤷‍♂️

dmex (approved):

Issue #1673, opened 2023-04-04
Crash in Windows view


MarekKnapek:
### Brief description of your issue

System Informer crashes.

### Steps to reproduce (optional)

 - Main menu.
 - View.
 - Windows.
 - Click on list view's "Modules" header.
 - Crash.

### Expected behavior (optional)

No crash.

### Environment (optional)

```shell
System Informer latest nightly.
```

dmex:
Duplicate of #1674

PR #1674, opened 2023-04-04
Fix crash when sorting by module in Windows view.

MarekKnapek:
https://github.com/winsiderss/systeminformer/issues/1673


dmex (changes requested):
PhCompareString just needs to be replaced with PhCompareStringWithNull

dmex (approved):

dmex:
I've updated your commits with the change for PhCompareStringWithNull 🥳https://github.com/winsiderss/systeminformer/pull/1674/commits/b1e8b2149f1e8b9e7b1093cc4b32df430041bd75

PR #1694, opened 2023-04-15
Fix UBs

namazso:


jxy-s:
@namazso, @dmex and I spoke offline on these changes. I haven't fully reviewed them and where the implications might lie elsewhere with these changes. But at a high level:

1. From @dmex - `0x3ff` to `0x3fful` changes the lock expressions and they evaluate to different values now, we will have to do a deeper review to fully understand the implications here.
2. The changes to locally copy the thunk data on to stack - it isn't clear to how this solves undefined behavior

Overall, please provide a description (or update the description of the PR) to include how/where you've identified the undefined behavior and why these changes address the undefined behavior.

namazso:
All of these were caught by compiling with clang-cl (after some minor warning-suppression and flag changes) using `-fsanitize=undefined -fsanitize-trap`. I kept fixing them until it didn't instantly crash on opening on my PC. Can't guarantee there aren't any more UBs.

> The changes to locally copy the thunk data on to stack - it isn't clear to how this solves undefined behavior

`memcpy` is guaranteed to copy bytes, so you can give it unaligned pointers. The copy on the stack is, of course, aligned correctly. Technically the cast itself earlier is UB, so that might be a good idea to just change into a PUCHAR too.

phlib/mapimg.c
```diff
@@ -1720,8 +1720,10 @@ NTSTATUS PhGetMappedImageImportDll(
             while (TRUE)
             {
                 PhpMappedImageProbe(ImportDll->MappedImage, entry, sizeof(IMAGE_THUNK_DATA32));
+                IMAGE_THUNK_DATA32 entry_copy;
+                memcpy(&entry_copy, entry, sizeof(IMAGE_THUNK_DATA32));
```
dmex, line 1724: The code is fine and there's no point using a memcpy here with MSVC.

phlib/fastlock.c
```diff
@@ -24,15 +24,15 @@
 #define PH_LOCK_EXCLUSIVE_WAKING 0x2
 
 #define PH_LOCK_SHARED_OWNERS_SHIFT 2
-#define PH_LOCK_SHARED_OWNERS_MASK 0x3ff
+#define PH_LOCK_SHARED_OWNERS_MASK 0x3fful
```
dmex, line 27: These types haven't changed in 14 years. This will change some of the lock expressions from signed to unsigned and I'm not sure if there will be side effects?

phlib/mapimg.c
namazso: MSVC does have an extension for unaligned pointers, but it's only if you use [__unaligned](https://learn.microsoft.com/en-us/cpp/cpp/unaligned?view=msvc-170). As far as standard C is concerned, this is currently UB.
I know on ARM (the non 64 bit) the original code would be terrible performance-wise as unaligned access is emulated by the kernel, but that architecture is fortunately dead and therefore won't ever be of any concern.
As with any UBs the real issue here is just that compilers are allowed to and will just do whatever. But granted, it does seem to function as intended on the current version of MSVC.

phlib/fastlock.c
namazso: The problem manifests here: https://github.com/winsiderss/systeminformer/blob/882e8adcb930e493a30d449d526679730dce226b/phlib/fastlock.c#L40
A shift that shifts into sign bit. C++20 changed it to IB, but we're doing C here.

The actual code usages work on ULONGs, so the issue doesn't happen there
https://github.com/winsiderss/systeminformer/blob/882e8adcb930e493a30d449d526679730dce226b/phlib/fastlock.c#L232-L238

phlib/mapimg.c
topia: (Just Curious) Because it's used only once, I doubt copying content is better than unaligned access. Marking as __unaligned is better than copying, isn't it?

dmex: Does this work? @namazso 

```
typedef IMAGE_THUNK_DATA32 UNALIGNED* UNALIGNED_PIMAGE_THUNK_DATA32;
UNALIGNED_PIMAGE_THUNK_DATA32 entry;

entry = (UNALIGNED_PIMAGE_THUNK_DATA32)ImportDll->LookupTable;
```

dmex (approved):

PR #1738, opened 2023-05-23
Fixes error C2198, too few arguments for call.

trparky:
Fixes error C2198, too few arguments for call.


trparky:
This isn't needed anymore as a recent commit by dmex fixed it.

dmex:
> This isn't needed anymore

Thanks for taking the time to fix the issue 👍 

Sorry about the timing, I had to merge two separate major changes that were dependant on each other using separate commits in different stages and that file wasn't updated while I fixed a bug with the second stage (deferred window layout).

PR #1740, opened 2023-05-28
Build CustomBuildTool.exe from source

antiufo:
As @RedKage [pointed out](https://github.com/winsiderss/systeminformer/issues/1668#issuecomment-1533162566), the `CustomBuildTool.exe` binary is currently committed to the repository.

Both for transparency / trust reasons, and to potentially help with antivirus detections, it might be useful to have CI build it from source.

There are a few more prebuilt binaries, but we can leave them for future PRs. `SystemInformer.sys` on the other hand is countersigned by Microsoft, and has zero AV detections (so no need to build it from source).


RedKage (commented):
👍

dmex:
de9f27d26b8fab86950a145c6a7e5c9a47209c8e

PR #1756, opened 2023-06-10
Fix setting creating thread id for process creation message

lstrsrt:
Unless I misunderstand the code (which is possible), this looks like a bug to me


lstrsrt:

jxy-s (approved):
deffinately a typo, thanks!

PR #1773, opened 2023-06-25
Fixed an issue where an error crash would occur when launching from t…

sonyps5201314:
…he accessibility features of the login screen.
Steps to reproduce:
1.turn off windows defender
2.Run cmd as an administrator, enter in cmd:
REG ADD "HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\utilman.exe" /t REG_SZ /v Debugger /d "C:\Program Files\SystemInformer\SystemInformer.exe" /f
3.lock screen.
4.Click Accessibility, and you will see the following error interface.
image

The patch provided by this PR fixes this issue.

### DIFF
--- a/plugins/UserNotes/db.c
+++ b/plugins/UserNotes/db.c
@@ -175,8 +175,10 @@ NTSTATUS LoadDb(
     NTSTATUS status;
     PVOID topNode;
     PVOID currentNode;
+    PH_STRINGREF sr;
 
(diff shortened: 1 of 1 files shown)


dmex:
PhGetStringRef would return an empty string which is a valid NT path. I've fixed the crash with https://github.com/winsiderss/systeminformer/commit/e9d1340e803e505b9dd5f2ce94e9777219ba93cd and added you as the coauthor 👍

PR #1817, opened 2023-08-20
A few minor fixes

DavidXanatos:
When digging through the protection code I noticed that the comment in KphpShouldSuppressObjectProtections seams to have been completely wrong so I made a new one.
And it seams KphApplyObProtections can be slightly streamlined...

### DIFF
--- a/KSystemInformer/protection.c
+++ b/KSystemInformer/protection.c
@@ -669,11 +669,9 @@ VOID KphApplyObProtections(
 
         if (KphpShouldPermitCreatorProcess(Info, actor, process))
         {
-            //
-            // add permissions which are not present in the usual KPH_PROTECED_PROCESS_MASK
-            //
-
-            allowedAccessMask |= (PROCESS_VM_OPERATION |
+            allowedAccessMask |= (KPH_PROCESS_READ_ACCESS |
+                                  PROCESS_TERMINATE |
+                                  PROCESS_VM_OPERATION |
                                   PROCESS_VM_WRITE);
 
             if ((allowedAccessMask & process->ProcessAllowedMask)
@@ -712,29 +710,26 @@ VOID KphApplyObProtections(
 
         allowedAccessMask = process->ThreadAllowedMask;
 
-        //if (KphpShouldPermitCreatorProcess(Info, actor, process))
-        //{
-        //    //
-        //    // add permissions which are not present in the usual KPH_PROTECED_THREAD_MASK
-        //    // currently none
-        //    //
-        //
-        //    //allowedAccessMask |= 
-        //
-        //    if ((allowedAccessMask & process->ThreadAllowedMask)
-        //        != allowedAccessMask)
-        //    {
-        //        KphTracePrint(TRACE_LEVEL_VERBOSE,
-        //                      PROTECTION,
-        //                      "Permitting extra thread handle access "
-        //                      "(0x%08x -> 0x%08x) in creator process %lu for "
-        //                      "process %lu",
-        //                      process->ThreadAllowedMask,
-        //                      allowedAccessMask,
-        //                      HandleToULong(actor->ProcessContext->ProcessId),
-        //                      HandleToULong(process->ProcessId));
-        //    }
-        //}
+        if (KphpShouldPermitCreatorProcess(Info, actor, process))
+        {
+            allowedAccessMask |= (KPH_THREAD_READ_ACCESS |
+                                  THREAD_TERMINATE |
+                                  THREAD_RESUME);
+
+            if ((allowedAccessMask & process->ThreadAllowedMask)
+                != allowedAccessMask)
+            {
+                KphTracePrint(TRACE_LEVEL_VERBOSE,
+                              PROTECTION,
+                              "Permitting extra thread handle access "
+                              "(0x%08x -> 0x%08x) in creator process %lu for "
+                              "process %lu",
+                              process->ThreadAllowedMask,
+                              allowedAccessMask,
+                              HandleToULong(actor->ProcessContext->ProcessId),
+                              HandleToULong(process->ProcessId));
+            }
+        }
 
         if ((desiredAccess & allowedAccessMask) != desiredAccess)
         {
--- a/KSystemInformer/protection.c
+++ b/KSystemInformer/protection.c
@@ -188,12 +188,12 @@ BOOLEAN KphpShouldSuppressObjectProtections(
     if (!NT_SUCCESS(status))
     {
         //
-        // If the process being protected doesn't dominate the target then
-        // do not strip any access from it.
+        // Grant access when the actor is a protected process and the target is
+        // not protected at a higher level.
         //
         KphTracePrint(TRACE_LEVEL_VERBOSE,
                       PROTECTION,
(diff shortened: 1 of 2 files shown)


KSystemInformer/protection.c
```diff
@@ -668,10 +668,12 @@ VOID KphApplyObProtections(
...
-        {
-            allowedAccessMask |= (KPH_PROCESS_READ_ACCESS |
-                                  PROCESS_TERMINATE |
-                                  PROCESS_VM_OPERATION |
+        {
+            //
+            // add permissions which are not present in the usual KPH_PROTECED_PROCESS_MASK
+            //
+
+            allowedAccessMask |= (PROCESS_VM_OPERATION |
```
jxy-s, line 676: I understand your intention for the change here but it conflicts with the design. While `ProcessAllowedMask` does in fact have what was removed set by way of `KPH_PROTECTED_PROCESS_MASK`, at this stage we are using a configured variable in the process context and this code should remain indifferent to what is set there. In other words, this code doesn't need knowledge of `KPH_PROTECTED_PROCESS_MASK`, and should remain that way.

There is no streamlining or savings of instructions here as the check for the creator still occurs along with masking-in of extra bits.

This design was chosen such that the reader doesn't need to know of external information. They need only read this block of code and understand that the creator process requires these permissions to function correctly. Furthermore, it minimizes the potential for mistake in the future - consider if the permissions in `KPH_PROTECTED_PROCESS_MASK` are adjusted or changed, this code would have to be known of and potentially adjusted too. The previous code functions all the same regardless of what is set. Now, with the change, if the code elsewhere changes a subtle bug might be introduced that could be overlooked.

An "optimal" implementation would do away with the masks inside of the context and rely only on the hard coded defines for a "verified" process. However, that design workflow severely restricts the extensibility of the code. A good example of this is when suppression of these protections was added when the machine is configured in debug mode (for developers). See this commit: https://github.com/winsiderss/systeminformer/commit/8ba24f6ef590b3ee58219286792349aa7541b5e8 - you'll see in that commit that the code here inside `protection.c` needn't change at all. At that time I had full confidence that making that change would not break this workflow downstream. With this change, I can no longer have full confidence that changing code elsewhere won't break this code.

KSystemInformer/protection.c
```diff
@@ -710,26 +712,29 @@ VOID KphApplyObProtections(
...
-                              "Permitting extra thread handle access "
-                              "(0x%08x -> 0x%08x) in creator process %lu for "
-                              "process %lu",
-                              process->ThreadAllowedMask,
-                              allowedAccessMask,
-                              HandleToULong(actor->ProcessContext->ProcessId),
-                              HandleToULong(process->ProcessId));
-            }
-        }
+        //if (KphpShouldPermitCreatorProcess(Info, actor, process))
```
jxy-s, line 715: This code's existence is intentional for the same reasons mentioned in the other comment.

Given `KPH_PROTECTED_THREAD_MASK` contains all the bits in that would be applied for the creator process, it's technically fine to remove this block - however, I stand firm on my principals for design here (as explained in the other comment). This would in fact save a call and some instructions. However, I firmly believe the design reasoning of the implementation outweighs the minuscule cost savings.

KSystemInformer/verify.c
```diff
@@ -356,7 +356,6 @@ NTSTATUS KphVerifyFile(
  * strict and lacks enough information from the kernel to fully understand the
  * protected process state.
  *
- * \param[in] Client - Client information.
```
jxy-s, line 359: 👍 - I'm also now noticing that the masks for protection are misspelled too

```diff
- KPH_PROTECED_PROCESS_MASK
+ KPH_PROTECTED_PROCESS_MASK
```

jxy-s (changes requested):

KSystemInformer/verify.c
jxy-s: I went ahead and removed this in this commit https://github.com/winsiderss/systeminformer/commit/5f0450fc5c876bca40a6181b42850bbfb86eb71c. I had another misc cleanup change inside `informer_debug.c`, and this was a good excuse to lump that change into it.

KSystemInformer/protection.c
```diff
@@ -186,14 +186,14 @@ BOOLEAN KphpShouldSuppressObjectProtections(
 
     status = KphDominationCheck(Target->EProcess, Actor->EProcess, UserMode);
     if (!NT_SUCCESS(status))
-    {
-        //
-        // If the process being protected doesn't dominate the target then
-        // do not strip any access from it.
+    {
+        //
+        // Grant access when the actor is a protected process and the target is
```
jxy-s, line 191: > I noticed that the comment in KphpShouldSuppressObjectProtections seams to have been completely wrong so I made a new one.

The comment and trace seems to read fine to me, what's the problem with leaving it?

KSystemInformer/protection.c
jxy-s: Also noteworthy, with this change, we lost some trace information - I think it's worth keeping the code the way it was for that reason too.

KSystemInformer/protection.c
DavidXanatos: The original comment says ´´´If the process being protected doesn't dominate the target then do not strip any access from it´´´
this means if booth processes have for example ```Protection.Type == PsProtectedTypeNone``` than no access will be stripped. But this is not what happens in such a scenario rightfully access is being stripped, as required for the protection to work.
What the code does is it allows any PPL protected process to access an KSI protected process.
So the original comment does not correctly reflect whats actually happening.

jxy-s: Okay 👍

jxy-s (approved):

DavidXanatos:
how about the above suggestion
> btw: tanking about trace information, always logging only the PID's os awfully inconvenient, since we have ImageFileName, how about always logging eider the entire path or even better picking out the filename itself and logging that.
> This could be made quite efficient by introducing a WCHAR* ImageName which would point into the buffer of ImageFileName.
>

shell I make a pull request for that?

jxy-s:
> how about the above suggestion
> 
> > btw: tanking about trace information, always logging only the PID's os awfully inconvenient, since we have ImageFileName, how about always logging eider the entire path or even better picking out the filename itself and logging that.
> > This could be made quite efficient by introducing a WCHAR* ImageName which would point into the buffer of ImageFileName.
> 
> shell I make a pull request for that?

I'd be fine with adding `ImageFileName` to some trace locations. Please note that `UNICODE_STRING` is not guaranteed to be null terminated. So using a `WCHAR*` that points to part of a `UNICODE_STRING` is errant. Is there anything wrong with using `%wZ` to just trace [`_KPH_PROCESS_CONTEXT.ImageFileName`](https://github.com/winsiderss/systeminformer/blob/master/KSystemInformer/include/kph.h#L1122)? If you just want the "file part", we could easily add a `UNICODE_STRING ImageFilePart;` to `_KPH_PROCESS_CONTEXT` to avoid tracing the entire path.

DavidXanatos:
The full path makes a log less readable and usually the filename itself is all that's needed to identify a process, I'll go with UNICODE_STRING ImageFilePart in _KPH_PROCESS_CONTEXT then.

Issue #1763, opened 2023-06-16
Kernel-mode driver in 3.0.6772 breaks Windows Sandbox


enervated:
After updating to 3.0.6772, I found Windows Sandbox totally unable to start so long as the kernel-mode driver had been loaded during the current Windows session. The errors thrown by Windows Sandbox vary widely, from null context errors, to network-related errors, and everything in between.

It's quite simple to reproduce: If at any point during the current Windows session KSystemInformer has loaded, Windows Sandbox will not launch.

Disable the driver, manually stop the driver (at which point it is deleted), then reboot. Windows Sandbox will then work.

Close Windows Sandbox. Run KSystemInformer by enabling the option in System Informer 3.0.6772. Windows Sandbox no longer launches.

I'm running Windows 10 22H2.

ge0rdi:
I can confirm the same behavior on Win10 22H2.

diversenok:
Also, it's not limited to Windows Sandbox but affects all Hyper-V-based virtual machines.

![image](https://github.com/winsiderss/systeminformer/assets/30962924/8ba0c0c9-bb6a-4a1b-905c-3852d5f5c5b2)

ge0rdi:
> The APC attempts to execute ntdll!RtlSetBits, triggers CFG, and crashes the process.

I think it is set up here:
https://github.com/winsiderss/systeminformer/blob/7c4fe0278b400822828cfd903e1e4155c41e2490/KSystemInformer/cid_tracking.c#L296

Though I have no clue why KPH is queueing such APC.
Hopefully SI authors will be able to shed more light on this.

enervated:
I saw vmcompute.exe crashes as well, although I did not dig into them to this extent. I did remove the CFG override that Windows automatically adds, though. This didn't fix the issue, it just changed the nature of the crashes.

Anyone investigating this issue on a system they care about should be aware that manually starting vmcompute.exe (i.e. the "Hyper-V Host Compute Service" service) to reproduce this bug is likely to cause a BSOD.

ge0rdi:
Thank you for the explanation.

> This should already be resolved with the next driver build.

Do you have any plans for the new build?

I'm also wondering why the problem doesn't disappear when driver is stopped. Only after reboot it is possible to start VM machines again.
Maybe there is also some other issue?

jxy-s:
[...]

PR #1832, opened 2023-08-27
Ignore CFG ES error while ACG is enabled

kkent030315:
I had a issue with between ACG mitigation policy and the CFG ES support implemented in below:

- https://github.com/winsiderss/systeminformer/issues/1763#issuecomment-1595935908

Applying ACG policy into SI (via process-specific settings in "Exploit Protection" for this case) causes unconnectable to driver because ACG will conflict with the CFG ES support feature.

`KphGuardGrantSuppressedCallAccess` call failure will lead to immediate unprotect and lack of `KPH_PROCESS_PROTECTED_PROCESS` process state.

Based on my investigation, there is no simple solution other than to either ignore the issue or use a hacky workaround.

```
 # Child-SP          RetAddr               Call Site
00 ffff8a83`d6fd3500 fffff806`15be5395     nt!MiCfgMarkValidEntries+0xb
01 ffff8a83`d6fd3510 fffff806`15a13ab5     nt!NtSetInformationVirtualMemory+0x495
02 ffff8a83`d6fd3880 fffff806`15a05f70     nt!KiSystemServiceCopyEnd+0x25
03 ffff8a83`d6fd3a88 fffff80b`bc46ea0c     nt!KiServiceLinkage
04 ffff8a83`d6fd3a90 fffff80b`bc45c9d3     SystemInformer!KphGuardGrantSuppressedCallAccess+0x8c [util.c @ 1390] 
05 ffff8a83`d6fd3b30 fffff80b`bc464ec2     SystemInformer!KphCheckProcessApcNoopRoutine+0x17b [cid_tracking.c @ 1906] 
06 ffff8a83`d6fd3ba0 fffff80b`bc464dc5     SystemInformer!KphpPerformProcessCreationTracking+0xaa [informer_process.c @ 446] 
07 ffff8a83`d6fd3c00 fffff806`15c4b39f     SystemInformer!KphpCreateProcessNotifyRoutine+0x25 [informer_process.c @ 543] 
08 ffff8a83`d6fd3c30 fffff806`15c452f2     nt!PspCallProcessNotifyRoutines+0x213
09 ffff8a83`d6fd3d00 fffff806`15c7b430     nt!PspInsertThread+0x68e
0a ffff8a83`d6fd3dc0 fffff806`15a13ab5     nt!NtCreateUserProcess+0xde0
0b ffff8a83`d6fd4a90 00007ffc`0ab0ea74     nt!KiSystemServiceCopyEnd+0x25
0c 000000ad`3bbde0c8 00007ffc`081c8e73     ntdll!NtCreateUserProcess+0x14
0d 000000ad`3bbde0d0 000000ad`3bbde6f0     0x00007ffc`081c8e73
0e 000000ad`3bbde0d8 00000000`00000000     0x000000ad`3bbde6f0
```

Please note that this is partial status of while single stepping so `nt!MiCfgMarkValidEntries+0xb` is not actually correct.
If my disassembly is right, this is just a sanity check and we can ignore it without any problems in my opinion.

Probably we can make a sanity check measure as well as how NT does, but that won't worth it compared to implementation cost so I decided to not do. However if anyone have greater idea than me my door is always open.

```c
__int64 __fastcall MiCfgMarkValidEntries(
        __int64 a1,
        __int64 a2,
        unsigned __int64 a3,
        _QWORD *a4,
        unsigned int a5,
        _DWORD *a6,
        char a7,
        __int64 a8,
        __int64 a9)
{
  // [redact]
  if ( v12 )
  {
    v17 = a4;
    while ( (*v17 & 0xFi64) == *(_DWORD *)(v14 + 0x18) )// 0x18=BitmapVad
    {
      v18 = v17[1];
      if ( (v18 & 0xFFFFFFFFFFFFFFF8ui64) != 0 )
        break;
      // (v18 & CFG_CALL_TARGET_VALID) != 0 is where this is triggered
      if ( (a7 & CFG_CALL_TARGET_VALID) != 0 && (v18 & CFG_CALL_TARGET_VALID) != 0
        || (v18 & CFG_CALL_TARGET_CONVERT_EXPORT_SUPPRESSED_TO_VALID) != 0
        && (!(unsigned int)MiIsProcessCfgExportSuppressionEnabled()
         || ((unsigned int)CfgAddressToBitState(v19 + a2, *(_QWORD *)v14) != 2
          || !(unsigned int)MiIsProcessCfgExportSuppressionEnabled())
         && !(unsigned int)MiValidateUserCallTarget(a2 + *v17, v14)) )
      {
        return 0xC0000022i64;
      }
      ++v16;
      v17 += 2;
      if ( v16 >= v12 )
        goto LABEL_14;
    }
    return 0xC000000Di64;
  }
  // [redact]
}
```

### DIFF
--- a/KSystemInformer/cid_tracking.c
+++ b/KSystemInformer/cid_tracking.c
@@ -1839,7 +1839,7 @@ NTSTATUS KphCheckProcessApcNoopRoutine(
 {
     NTSTATUS status;
     HANDLE processHandle;
-    PROCESS_MITIGATION_POLICY_INFORMATION policyInfo;
+    PROCESS_MITIGATION_POLICY_INFORMATION policyInfo, policyInfo2;
     ULONG flags;
 
     PAGED_CODE_PASSIVE();
@@ -1883,6 +1883,22 @@ NTSTATUS KphCheckProcessApcNoopRoutine(
         goto Exit;
     }
 
+    policyInfo2.Policy = ProcessDynamicCodePolicy;
+    policyInfo2.DynamicCodePolicy.Flags = 0;
+    status = ZwQueryInformationProcess(processHandle,
+                                       ProcessMitigationPolicy,
+                                       &policyInfo2,
+                                       sizeof(policyInfo2),
+                                       NULL);
+    if (!NT_SUCCESS(status))
+    {
+        KphTracePrint(TRACE_LEVEL_ERROR,
+                      TRACKING,
+                      "ZwQueryInformationProcess failed: %!STATUS!",
+                      status);
+        goto Exit;
+    }
+
     //
     // TODO(jxy-s) investigate any needed support for XFG
     //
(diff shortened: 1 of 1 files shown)


jxy-s:
@kkent030315 sorry for the delayed response here. First, thank you for taking the time to investigate and put together a PR. I appreciate the thorough information and proposed solution.

What you have would avoid the immediate problem. But it is, as you said, a "hacky workaround". ACG support is an oversight (my fault) with the current implementation. And yes, the kernel does block the ability to modify the CFG bitmap when ACG is turned on. However, that doesn't mean they are mutually exclusive mitigations. You can have ACG and CFG on at the same time.

There is a better solution here. There is a feature of ACG that enable a thread to be marked as "opt-out" and modify the CFG bitmask. I'm currently investigating this avenue to provide ACG support.

Finally, I want to also note that there are other ares of SI that are incompatible with ACG. That said, we should still address this incompatibility with the KPH mitigations. I've created this issue for tracking (https://github.com/winsiderss/systeminformer/issues/1833) I will address this problem myself once I finish my investigation.

I won't be able to accept this solution as is.

PR #1849, opened 2023-09-16
Fix compile error.

trparky:
Fixes issue that results in this ugly error message.

[ERROR] (1) C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(456,5): error MSB8020: The build tools for WindowsApplicationForDrivers10.0 (Platform Toolset = 'WindowsApplicationForDrivers10.0') cannot be found. To build using the WindowsApplicationForDrivers10.0 build tools, please install WindowsApplicationForDrivers10.0 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [C:\Users\trpar\source\repos\systeminformer\sistatus\sistatus.vcxproj]

### DIFF
--- a/sistatus/sistatus.vcxproj
+++ b/sistatus/sistatus.vcxproj
@@ -35,13 +35,13 @@
     <Platform Condition="'$(Platform)' == ''">Win32</Platform>
     <RootNamespace>sistatus</RootNamespace>
     <DriverType>KMDF</DriverType>
-    <WindowsTargetPlatformVersion>$(LatestTargetPlatformVersion)</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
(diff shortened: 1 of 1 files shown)


jxy-s:
@trparky you're missing the WDK: https://learn.microsoft.com/en-us/windows-hardware/drivers/download-the-wdk

With the change you're suggesting, building the project is broken since the `.mc` artifacts are not rebuilt when changes happen. Let me consider reorganizing the projects so those who don't want to build the kernel parts don't have to have the WDK.

trparky:
OK.

trparky:
Yeah, for those that don't have a signing certificate to sign the kernel driver, there's no point in compiling the kernel part and thus we use what's in the repo.

jxy-s:
I understand. The status messages are not just for the kernel parts. There is nothing stopping you from installing the WDK and never building a kernel component - I'd recommend having it installed anyway, since there's stuff defined there that isn't elsewhere 😁 .

Regardless, I understand your point of view. I can reorganize things dmex and I can maintain and manage this without causing undue environmental impact of the WDK on others 😉.

trparky:
Alrighty then, I look forward to your commit.

jxy-s:
See commit: https://github.com/winsiderss/systeminformer/commit/4bc8f19d3376db35455d99d5ad2dab360450edb3

PR #1857, opened 2023-10-01
phnt tabs to spaces

m417z:
You might want to add a check for it, should be easy to do with GitHub Actions.

Also, fixed a small function declaration inconsistency.


dmex:
> You might want to add a check for it, should be easy to do with GitHub Actions.

We have checks for this with editorconfig:
https://github.com/winsiderss/systeminformer/blob/master/.editorconfig

 VSCode just decides to ignore it 🤷‍♂️

dmex:
I've merged this as co-author from e7096f72b4374a1c24f69096667df0aa3e026790 👍

PR #1877, opened 2023-10-16
SourceLink support

ge0rdi:
This PR adds support for SourceLink so that debuggers can obtain proper version of source files automatically.
It uses official `Microsoft.SourceLink.GitHub` NuGet package that does all the work and is referenced by all projects.

Additionally, it fixes paths stored in PDBs so that they are all relative to repo root.
Currently these paths are solution-relative so it is not possible to map them to repo paths for some binaries (`systeminformer.sys`, plugins, ...).


tools/CustomBuildTool/bin/Release/CustomBuildTool.exe
ge0rdi, line 1: I have built custom build tool using `build_tools.cmd`. But maybe you'd like to rebuild it on your own?

ge0rdi:
It will be necessary to also rebuild `thirdparty.lib` because current one contains source paths relative to `tools/thirdparty` (such as `pcre/pcre2_compile.c`) and these don't match SourceLink data.

jxy-s:
Just finished testing, looks to work as advertised, nice work! 👍

tools/CustomBuildTool/bin/Release/CustomBuildTool.exe
jxy-s: I'll take care of this post-merge.

jxy-s (approved):

PR #1878, opened 2023-10-17
Update ntrtl.h

elnx:
According to [this MSDN documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-barrierafterread), `BarrierAfterRead` is not available until win11 22h2.

### DIFF
--- a/phnt/include/ntrtl.h
+++ b/phnt/include/ntrtl.h
@@ -1257,7 +1257,7 @@ RtlWakeAddressSingle(
 
 // end_rev
 
-#if PHNT_VERSION >= PHNT_WIN11_22H2
+#if (PHNT_VERSION >= PHNT_WIN11_22H2)
 FORCEINLINE
 VOID
 RtlCopyVolatileMemory(
--- a/phnt/include/ntrtl.h
+++ b/phnt/include/ntrtl.h
@@ -1257,6 +1257,7 @@ RtlWakeAddressSingle(
 
 // end_rev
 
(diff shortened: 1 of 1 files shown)


elnx:

phnt/include/ntrtl.h
```diff
@@ -1257,6 +1257,7 @@ RtlWakeAddressSingle(
 
 // end_rev
 
+#if PHNT_VERSION >= PHNT_WIN11_22H2
```
dmex, line 1260: This needs brackets `(PHNT_VERSION >= PHNT_WIN11_22H2)`

dmex (changes requested):

Biswa96:
But BarrierAfterRead is just an inline function with `__dmb` intrinsic in wdm.h. Does it depend on Windows version?

elnx:
> But BarrierAfterRead is just an inline function with `__dmb` intrinsic in wdm.h. Does it depend on Windows version?

You are right. The newer Windows SDK (10.0.22621.0) has defined it as an inline function, in winnt.h, while there is no definitions for it in older SDK (for example 10.0.19041.0).

Biswa96:
> while there is no definitions for it in older SDK (for example 10.0.19041.0).

Latest Windows SDK is already an requirement for other functions. IIRC, there were some issues where using latest SDK fixes compiler errors. Please correct me if I am wrong.

dmex (approved):

Issue #1881, opened 2023-10-23
Unable to locate the file.


MarekKnapek:
Steps to reproduce:

 - Windows Server 2022.
 - System Informer 3.0.7270 (d44e4d0).
 - Create a *.VHD hard disk image file, attach it, create a volume in it.
 - Mount the volume to NTFS path, do not use drive letter.
 - Copy some executable into the volume.
 - Run the executable.
 - Find process of the new executable in System Informer, open its properties.
 - Click the magnifying glass button next to path to the executable.

Expected result:
 - PE inspector tool appears.

Actual result:
 - Error message about file not found appears.

![pe](https://github.com/winsiderss/systeminformer/assets/13893067/058d44d9-b585-4a68-9268-1bb2bac29a3f)

[Window Title]
System Informer

[Main Instruction]
Unable to locate the file.

[Content]
The object was not found.

[Close]

PR #1883, opened 2023-10-24
Fix for #1881, could not open PE viewer.

MarekKnapek:
Use Win32 name from ProcessImageFileNameWin32 instead of NT name from SystemProcessIdInformation in case NT to Win32 name conversion fails.


dmex:
PR was outdated by https://github.com/winsiderss/systeminformer/commit/8523f836f7496016a7809582fa987e325ce90350

> Use Win32 name from ProcessImageFileNameWin32 instead of NT name

Win32 names are deprecated - the code for handling cases like #1881 has been in the repo for a while but is still being tested and not enabled at the moment.

PR #1885, opened 2023-10-28
Unconditionally include ntkeapi.h in phnt.h

mrexodia:
Without this the following definitions are missing in kernel mode:
- KTHREAD_STATE
- KHETERO_CPU_POLICY
- LDR_*

The `ntkeapi.h` already does checks on the `PHNT_MODE`, so this should work in all cases.

I tested this locally in user mode by doing:

```c
#include <phnt_windows.h>
#include <phnt.h>
```

And kernel mode:

```cpp
extern "C"
{
#include <ntifs.h>
#include <ntddstor.h>
#include <mountdev.h>
#include <ntddvol.h>
#include <ntstrsafe.h>
#include <ntimage.h>
#include <wdm.h>
}

#include <phnt.h>
```

### DIFF
--- a/phnt/include/phnt.h
+++ b/phnt/include/phnt.h
@@ -74,9 +74,9 @@ extern "C" {
 #if (PHNT_MODE != PHNT_MODE_KERNEL)
 #include <phnt_ntdef.h>
 #include <ntnls.h>
-#include <ntkeapi.h>
 #endif
 
+#include <ntkeapi.h>
 #include <ntldr.h>
 #include <ntexapi.h>
 


mrexodia:

dmex:
ntkeapi is already included for both user and kernel?

mrexodia:
As far as I can tell it’s not included in kernel mode 🤷‍♂️ If it is perhaps it’s too early because without this change my driver doesn’t compile…

dmex:
We're importing with KSI:

https://github.com/winsiderss/systeminformer/blob/2c9dd8765011a85fd98774cb37ff4ef52923d8c1/KSystemInformer/include/kph.h#L12-L28

mrexodia:
I did some more digging and it turns out this happens because of the amalgamation I'm doing. There is `#include <ntexapi.h>`, which has `#include <ntkeapi.h>` right at the top of it. My patch doesn't affect the compilation of KSystemInformer (or SystemInformer), it just includes `ntkeapi.h` a little bit earlier.

Would be appreciated if you could merge this, because working around this downstream is going to be annoying and it looks like the intention was to include `ntkeapi.h` in both user and kernel mode anyway...

jxy-s:
@mrexodia I think you change is reasonable here. There is some other cleanup I want to do for better kernel support in general.

jxy-s (approved):

dmex (approved):

jxy-s:
Manually merged here: https://github.com/winsiderss/systeminformer/commit/0a23ee7ea539af23c5a9d71aee5a6445e0745e8b

PR #1899, opened 2023-11-15
Add Efficient/Performance core querying on ARM64

hmartinez82:


SystemInformer/syssccpu.c
```diff
@@ -1822,10 +1822,78 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+
+    for (
+        processorInfo = logicalInformation;
+        (ULONG_PTR)processorInfo < (ULONG_PTR)PTR_ADD_OFFSET(logicalInformation, bufferLength);
+        processorInfo = PTR_ADD_OFFSET(processorInfo, processorInfo->Size)
+        )
+    {
+        //One logical core per bit set in the core mask
+        ULONG_PTR mask = processorInfo->Processor.GroupMask[0].Mask;
+        while (mask)
```
hmartinez82, line 1848: Is there a better way to do this?

jxy-s: I think this will give you what you want and leverage intrinsic. I believe the intrinsic is supported everywhere?

```c
    ULONG index;

    _BitScanForward(&index, mask);

    mask &= ~(1UL << index);
```

SystemInformer/syssccpu.c
```diff
@@ -1822,10 +1822,78 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+    }
+
+    if (!hybridProcessorTypeList || ProcessorIndex > hybridProcessorTypeList->Count)
+        return NULL;
+
+    switch (PtrToUlong(hybridProcessorTypeList->Items[ProcessorIndex]))
+    {
+    case 0:
+        {
+            static PH_STRINGREF hybridECoreTypeSr = PH_STRINGREF_INIT(L"E-Core");
```
jxy-s, line 1887: ARM names this `LITTLE` in `big.LITTLE` mixed cores.

SystemInformer/syssccpu.c
```diff
@@ -1822,10 +1822,78 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+            return &hybridECoreTypeSr;
+        }
+    case 1:
+        {
+            static PH_STRINGREF hybridPCoreTypeSr = PH_STRINGREF_INIT(L"P-Core");
```
jxy-s, line 1892: Likewise this would be the `big` in `big.LITTLE`.

jxy-s (changes requested):

SystemInformer/syssccpu.c
hmartinez82: Aw, ok, the bit flipping will require less loops because on average the mask become 0 with fewer iterations than scanning all the 64 bits for 1s and 0s.

jxy-s: > average the mask become 0 with fewer iterations than scanning all the 64 bits for 1s and 0s

Ah, I see what you're doing here. We use `PhCountBitsUlongPtr` elsewhere which is effectively a popcount, which should be faster than the bit twiddling here?

hmartinez82: Oh. Nice, akin to https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlnumberofsetbitsulongptr ? (To be frank I never heard about this one either :) First time I'm seeing it)

hmartinez82: PhCountBitsUlongPtr(), superb, it does exactly what I wanted. I also fixed the previous code since it was only considering the first group mask.

SystemInformer/syssccpu.c
```diff
@@ -1822,10 +1822,80 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+        PhInitializeHybridProcessorTypeCache(&hybridProcessorTypeList);
+        PhEndInitOnce(&initOnce);
+    }
+
+    if (!hybridProcessorTypeList || ProcessorIndex > hybridProcessorTypeList->Count)
+        return NULL;
+
+    switch (PtrToUlong(hybridProcessorTypeList->Items[ProcessorIndex]))
+    {
+    case 0:
```
jxy-s, line 1887: Does it make sense to have a define for these, like:
```c
#define ARM_CORETYPE_LITTLE 0
#define ARM_CORETYPE_BIG     1
```

jxy-s:
I'm curious if there is a similar method to what we do for x86/AMD64. I wonder if there is system register we can read with `MRS` on the current thread. Or conversely, if the approach you have here has any advantages where we could consider doing it for on x86/AMD64 too?

The documentation on `EfficiencyClass` sparse and from what I can tell the query information retrieves this information here:

```c
CHAR __fastcall KeGetActualProcessorEfficiencyClass(
        _KPRCB *ProcessorControlBlock)
{
    if ( KeHeteroSystem )
    {
        return ProcessorControlBlock->PowerState.ArchitecturalEfficiencyClass;
    }
    else
    {
        return ProcessorControlBlock->PowerState.EarlyBootArchitecturalEfficiencyClass;
    }
}
```

The early boot looks to be propagated from the initial `KPRCB`. I haven't been able to locate the alternative once `KeHeteroSystem` is identified in the boot process.

I'm a little concerned there might be an unforeseen issue with grouping or other boot configurations which might report inaccurate information. It is probably ideal for us to ask the processor directly of the `big.LITTLE` state/support - as we do on x86/ARM64. Here is an example of us making emitting `MSR` to read system register state:

https://github.com/winsiderss/systeminformer/blob/5d97d6b3f99bd7c651b448ae414f39150cf9af2f/SystemInformer/syssccpu.c#L1313-L1316

I wonder if there is a similar `MRS` for reading the `big.LITTLE` state?

hmartinez82:
I once found a register that can give this information, but it's been more than an year and I remember not being possible to read that register from userspace code because it was EL1

SystemInformer/syssccpu.c
hmartinez82: Done.
Also, how do I sign the CLA?

hmartinez82:

hmartinez82:
recheck

jxy-s (approved):

PR #1902, opened 2023-11-16
Add ARM64 exceptions

hmartinez82:
Initial discussion for something better than a blank page :)
Is this acceptable? 

image


hmartinez82:
Also, a very newbie question: How do I read the `.xdata` section? I've been scouring the internet, but I can't for the life of me find how to read it!

jxy-s:
Does `ARM64EC` have a table we need to locate and extract alternate exception info from?

> Also, a very newbie question: How do I read the .xdata section? I've been scouring the internet, but I can't for the life of me find how to read it!

`UnwindData` I believe is an RVA, you can use `PhMappedImageRvaToSection` to locate the section. The RVA to a virtual address with `PhMappedImageRvaToVa`.

hmartinez82:
@jxy-s I don't fully understand how to use `PhMappedImageRvaToSection()` in this case. If I understand correctly I would need the RVA of where .xdata starts and use `PhMappedImageRvaToSection()`, right? But I don't know that value wither. The `UnwindData` field in each exception entry is the RVA inside `.xdata`. At least that's what I understood from:

_Exception Information RVA is the address of the variable-length exception information structure, stored in the .xdata section. This data must be 4-byte aligned._

jxy-s:
> UnwindData is the Exception Information RVA

`PhMappedImageRvaToSection` takes an RVA and returns the `IMAGE_SECTION_HEADER` for which that RVA resides. If you want the information about that section you can use `PhMappedImageRvaToSection` to get it. You have an RVA, it's `UnwindData`. That should get you to the `.xdata` `IMAGE_SECTION_HEADER`. But, if you want to just read what information is at `UnwindData` you can use `PhMappedImageRvaToVa` which will look up the section in which it resides and do the calculation to give you a pointer to the data in the section mapped via the `PH_MAPPED_IMAGE` structure.

So, since I think your goal is to read the exception information stored at the `UnwindData` RVA, you should just `PhMappedImageRvaToVa` to get the VA into the section mapping.

hmartinez82:
You are right. Turns out I don't need to know where `.xdata` starts. `PhMappedImageRvaToVa()` alone did the trick:
image

hmartinez82:
Regarding ARM64EC. I have no idea. I can work on that next 😅

jxy-s:
Some of these `ARM64` items were left unimplemented simply because I had lost the energy to keep working on it - `ARM64EC` is a beast... we have support for the it in most places. In general, I was looking for time to find the right way to approach some of outstanding `ARM64` items. I appreciate your time and effort helping to move this along ❤️

DavidXanatos:
for ARM64EC you may want to review this project: https://github.com/FFRI/ProjectChameleon

tools/peview/peexceptprp.c
```diff
@@ -166,7 +166,80 @@ VOID PvEnumerateExceptionEntries(
...
+            WCHAR value[PH_INT64_STR_LEN_1];
+            IMAGE_ARM64_RUNTIME_FUNCTION_ENTRY_XDATA* exceptionData;
+
+            PhPrintUInt64(value, ++count);
+            lvItemIndex = PhAddListViewItem(ListViewHandle, MAXINT, value, entry);
+
+            PhPrintPointer(value, UlongToPtr(entry->BeginAddress));
+            PhSetListViewSubItem(ListViewHandle, lvItemIndex, 1, value);
+
+            ULONG functionLength = entry->FunctionLength << 2;
```
jxy-s, line 184: I think we should move accessing the packed `FunctionLength` into the case where we know the entry is in the packed form. Especially since I'm asking for a null pointer check elsewhere.

tools/peview/peexceptprp.c
```diff
@@ -166,7 +166,80 @@ VOID PvEnumerateExceptionEntries(
...
+            if (entry->Flag != 0) //Packed Unwind Data
+            {
+                PhPrintPointer(value, UlongToPtr(entry->UnwindData));
```
jxy-s, line 187: Accessing `UnwindData` here is incorrect. This is an erroneous value when `Flags != 0`.

tools/peview/peexceptprp.c
```diff
@@ -166,7 +166,80 @@ VOID PvEnumerateExceptionEntries(
...
+            IMAGE_ARM64_RUNTIME_FUNCTION_ENTRY_XDATA* exceptionData;
+
+            PhPrintUInt64(value, ++count);
+            lvItemIndex = PhAddListViewItem(ListViewHandle, MAXINT, value, entry);
+
+            PhPrintPointer(value, UlongToPtr(entry->BeginAddress));
+            PhSetListViewSubItem(ListViewHandle, lvItemIndex, 1, value);
+
+            ULONG functionLength = entry->FunctionLength << 2;
+            if (entry->Flag != 0) //Packed Unwind Data
```
jxy-s, line 185: Is it worth showing the type of packing here (instead of yes/no), the bits are described here:
https://learn.microsoft.com/en-us/cpp/build/arm64-exception-handling?view=msvc-170#packed-unwind-data

tools/peview/peexceptprp.c
jxy-s: It's probably best to instead show a representation of the remaining unrepresented bits here:
```c
            ULONG RegF : 3;
            ULONG RegI : 4;
            ULONG H : 1;
            ULONG CR : 2;
            ULONG FrameSize : 9
```

Described here: https://learn.microsoft.com/en-us/cpp/build/arm64-exception-handling?view=msvc-170#packed-unwind-data

jxy-s (changes requested):
Apologies on the delay here. I just got around to testing the changes myself. I've added some feedback to incorporate.

Additionally there is `AMD64` in the `ARM64EC` binary (you will see then when you use `dumpbin.exe`). I haven't located where they are. If I had to guess the dynamic relocations fix up the `IMAGE_DIRECTORY_ENTRY_EXCEPTION` directory in the headers to point to the alternative directory. The loader will do this for `ARM64EC` in the case of the export directory, see `PhpFixupExportDirectoryForARM64EC` for the hoops necessary to locate the alternate directory.

tools/peview/peexceptprp.c
jxy-s: To be clear, instead of showing `L"Yes"` for "Packed", we could instead show something more descriptive:
```
    00 = packed unwind data not used; remaining bits point to an .xdata record
    01 = packed unwind data used with a single prolog and epilog at the beginning and end of the scope
    10 = packed unwind data used for code without any prolog and epilog. Useful for describing separated function segments
    11 = reserved.
```

tools/peview/peexceptprp.c
jxy-s: To be clear, instead of showing L"Yes" for "Packed", we could instead show something more descriptive:

```
    00 = packed unwind data not used; remaining bits point to an .xdata record
    01 = packed unwind data used with a single prolog and epilog at the beginning and end of the scope
    10 = packed unwind data used for code without any prolog and epilog. Useful for describing separated function segments
    11 = reserved.
```

jxy-s:
Yep the loader does the same thing with the exception directory that it does with the export directory, see this:
image

We're looking at the headers of the `ARM64EC` `ntdll.dll`. On the left is a native `ARM64` binary. On the right is an `AMD64` binary running in emulation mode.

`ARM64`:
```
  32A8A0 [   15BC9] address [size] of Export Directory
       0 [       0] address [size] of Import Directory
  38C000 [   75070] address [size] of Resource Directory
  352000 [   12D18] address [size] of Exception Directory
  3EE600 [    B4F0] address [size] of Security Directory
  402000 [     E70] address [size] of Base Relocation Directory
  2CAAD0 [      54] address [size] of Debug Directory
       0 [       0] address [size] of Description Directory
       0 [       0] address [size] of Special Directory
       0 [       0] address [size] of Thread Storage Directory
  2CA990 [     140] address [size] of Load Configuration Directory
       0 [       0] address [size] of Bound Import Directory
       0 [       0] address [size] of Import Address Table Directory
       0 [       0] address [size] of Delay Import Directory
       0 [       0] address [size] of COR20 Header Directory
       0 [       0] address [size] of Reserved Directory
```

`x64`:
```
  314C30 [   15C70] address [size] of Export Directory
       0 [       0] address [size] of Import Directory
  38C000 [   75070] address [size] of Resource Directory
  364D18 [    177C] address [size] of Exception Directory
  3EE600 [    B4F0] address [size] of Security Directory
  402000 [     E70] address [size] of Base Relocation Directory
  2CAAD0 [      54] address [size] of Debug Directory
       0 [       0] address [size] of Description Directory
       0 [       0] address [size] of Special Directory
       0 [       0] address [size] of Thread Storage Directory
  2CA7D0 [     140] address [size] of Load Configuration Directory
       0 [       0] address [size] of Bound Import Directory
       0 [       0] address [size] of Import Address Table Directory
       0 [       0] address [size] of Delay Import Directory
       0 [       0] address [size] of COR20 Header Directory
       0 [       0] address [size] of Reserved Directory
```

So, to support showing the `AMD64` exports in a `ARM64EC` binary - as we do for the exports already. You will want to use a similar workflow as we for the export directory lookup in `PhpFixupExportDirectoryForARM64EC`.

tools/peview/peexceptprp.c
```diff
@@ -166,7 +166,80 @@ VOID PvEnumerateExceptionEntries(
     }
     else if (imageMachine == IMAGE_FILE_MACHINE_ARM64)
     {
-        /* Todo */
+        for (ULONG i = 0; i < exceptions.NumberOfEntries; i++)
+        {
+            PPH_IMAGE_RUNTIME_FUNCTION_ENTRY_ARM64 entry = PTR_ADD_OFFSET(exceptions.ExceptionDirectory, i * sizeof(PH_IMAGE_RUNTIME_FUNCTION_ENTRY_ARM64));
```
jxy-s, line 171: `IMAGE_RUNTIME_FUNCTION_ENTRY_ARM64` is in the SDK headers now, I think we can eliminate the `PPH_IMAGE_RUNTIME_FUNCTION_ENTRY_ARM64`?

hmartinez82:
Tank you for the many change requests. I'll work on the ones that are up to my league once I'm back after December, 11th

tools/peview/peexceptprp.c
hmartinez82: Any idea on column names and what text for each case?

jxy-s:
@hmartinez82 I'm currently working on a set of changes specific to ARM to provide better support and visibility. I wanted to ask if you would like me to incorporate your changes here to that set of changes, or if you plan to continue this PR?

hmartinez82:
Sure, you can incorporate these changes in your PR and I'll close this one :)

PR #1910, opened 2023-11-25
Fix VS_FIXEDFILEINFO.dwFileOS to string conversion

alabuzhev:
The implementation erroneously assumed that dwFileOS is a set of flags, but actually it contains two separate enumerations in high and low words.

As a result, peview was showing rubbish for most of the currently encountered binaries, e.g. "DOS, Windows NT (0x40004)" for kernel32.dll from Win11 x64, which has nothing to do with DOS.

Reimplemented in accordance with [VS_FIXEDFILEINFO](https://learn.microsoft.com/en-us/windows/win32/api/verrsrc/ns-verrsrc-vs_fixedfileinfo).


alabuzhev:

dmex (approved):

dmex:
LGTM good catch 👍

PR #1921, opened 2023-12-02
Fix another ALPC parameter type error and add support for vs2019.

kanren3:
I missed another ALPC parameter type error, and hope that phnt can continue to support vs2019.


Biswa96:
Would it be possible to separate the changes into two commits - one for the fix and another one for the vs2019 support?

kanren3:
Ok, i will break up the commit

PR #1964, opened 2024-01-15
Fix DEP mitigation reporting

diversenok:
This pull request fixes a bug that makes System Informer report the opposite state of the DEP mitigation for processes that don't have permanent DEP.


phlib/native.c
```diff
@@ -1486,9 +1486,9 @@ NTSTATUS PhGetProcessDepStatus(
 
     // Check if execution of data pages is enabled.
     if (executeFlags & MEM_EXECUTE_OPTION_ENABLE)
-        depStatus = 0;
-    else
         depStatus = PH_PROCESS_DEP_ENABLED;
+    else
+        depStatus = 0;
```
jxy-s, line 1491: Nice find 👍 

My only question is if `MEM_EXECUTE_OPTION_ENABLE` and `MEM_EXECUTE_OPTION_DISABLE` are truly mutually exclusive flags. I haven't looked at this too deeply, but it seems like in `KeSetExecuteOptions` they could both end up being set? I have not looked at the locations where these are checked, but if `DISABLE` takes precedence over `ENABLE` then we should also be checking for that. We should double check that and correct as necessary.

Regardless, this is certainly an acceptable bugfix. I will merge as is for now, pending confirmation from the above.

jxy-s (approved):

phlib/native.c
jxy-s: I took a brief look at `KeGetExecuteOptions` and it seems like `ENABLE` takes precedence over `DISABLE`. So I believe what we have here is correct.

dmex (commented):
✔️

PR #1975, opened 2024-02-02
Fix Protection column for IUM processes

ge0rdi:
Protection column will show information for Isolated User Mode (IUM) processes now.

### DIFF
--- a/SystemInformer/proctree.c
+++ b/SystemInformer/proctree.c
@@ -3515,7 +3515,7 @@ BOOLEAN NTAPI PhpProcessTreeNewCallback(
                 break;
             case PHPRTLC_PROTECTION:
                 {
-                    if (processItem->Protection.Level != 0 && processItem->Protection.Level != UCHAR_MAX)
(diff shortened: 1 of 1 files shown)


ge0rdi:
`LsaIso.exe` is an example of IUM protected process.
SI shows protection in General tab, but Protection column in process tree is empty.

jxy-s (approved):
I'm okay with this. An alternative would be adding another column for `Secure (IUM)` - @dmex thoughts?

dmex:
> adding another column for Secure (IUM)

I didn't add a separate column because there's only 2 processes and not enough to justify a column 🤔

dmex (approved):

jxy-s:
> > adding another column for Secure (IUM)
> 
> I didn't add a separate column because there's only 2 processes and not enough to justify a column 🤔

Yeah, agreed. Just wanted to double check on this with you. I also prefer the consistency between the properties and the column with respect to "Protection". 👍

PR #1978, opened 2024-02-07
Show ARM64EC vs ARM64X in PEView's Target Machine

hmartinez82:
I was investigating the Image Type of Winword.exe, Excel.exe etc and I noticed that the Windows on ARM binaries were being shown as AMD64. But it's truly ARM64EC.
Using `dumpbin` on the three types of ARM64 binaries I found these outputs:

Pure ARM64:
```
FILE HEADER VALUES
            AA64 machine (ARM64)
```

ARM64EC:
```
FILE HEADER VALUES
            8664 machine (x64) (ARM64X)
```

ARM64X:
```
FILE HEADER VALUES
            AA64 machine (ARM64) (ARM64X)
```

It wouldbe nice if that ARM64X was not being repeated in the ARM64EC binaries, but I digress.

Please correct me if I'm wrong or if there's a less cheesy way to tell the three types of ARM64 binaries apart, but all I could do was to check for the presence of the `.hexpthk` in combination with the IMAGE TYPE in the PE header leading to this truth table:

| PE Header -> | AMD64 | ARM64 |
|--------------|:-----:|:-----------:|
| w/o `.hexpthk` |  AMD64 | ARM64 |
| w/ `.hexpthk` |  ARM64EC | ARM64X |

### DIFF
--- a/tools/peview/peprp.c
+++ b/tools/peview/peprp.c
@@ -867,11 +867,24 @@ VOID PvpSetPeImageMachineType(
 {
     ULONG machine;
     PWSTR type;
+    size_t sz;
+    size_t szComp = strlen(".hexpthk");
+    boolean hasA64ThunkSection = FALSE;
 
     if (PvMappedImage.Magic == IMAGE_NT_OPTIONAL_HDR32_MAGIC)
         machine = PvMappedImage.NtHeaders32->FileHeader.Machine;
     else
         machine = PvMappedImage.NtHeaders->FileHeader.Machine;
+
+    for (ULONG i = 0; i < PvMappedImage.NumberOfSections; i++)
+    {
+        sz = min(strlen(PvMappedImage.Sections[i].Name), szComp);
+        if (strncmp(PvMappedImage.Sections[i].Name, ".hexpthk", sz) == 0)
+        {
+            hasA64ThunkSection =  TRUE;
+            break;
+        }
+    }
 
     switch (machine)
     {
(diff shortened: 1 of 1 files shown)


jxy-s:
@hmartinez82 I actually have another local branch working on a variety of other ARM64 changes. And I already have a similar change on it.

Much of the office suite is actually `x64 (ARM64X)` and there are some others (like dllhost) which is `ARM64 (ARM64X)`. So Let's hold off on the changes you have here please.

I will merge those changes when they're ready. They also contain changes to show the "Architecture" in the modules list, along with a number of other quality of life improvements around ARM.

The stack walking in ARM is also a bit broken for certain cases around push thunk and enter/exit thunks. I'm working on resolving that too.

I will also be touching base in your other PR shortly.

hmartinez82:
My whole point is that `x64 (ARM64X)` is actually ARM64EC
and `ARM64 (ARM64X)` is ARM64X

That was what I'm explaining in the PR details. Sorry if it wasn't clear.

jxy-s:
Yeah we're talking past one another. I think that's expected given the cluster that is Windows ARM64 terminology....

`ARM64EC` is just an ABI for `ARM64`... the PE formats for `x64 (ARM64X)`, `ARM64 (ARM64X)`, and `x86 (CHPE)` _all_ contain `ARM64EC` code... `ARM64X` and `CHPE` is a hybrid binary format for a given architecture. 

And it gets worse depending on where you look... Microsoft uses `CHPEv2` (`dbgeng.dll`) and `ARM64X` (`dumpbin.exe` / `link.exe`) interchangeably. But `CHPE` is the x86 terminology... It seems like Microsoft is converging on `ARM64X` for a 64-bit binary and `CHPE` for 32-bit.

Anyway... `ARM64EC` isn't a hybrid executable format... it's an ARM64 ABI that exists inside the hybrid binaries, the JIT code in a process running under the emulator, or the cached JIT `XtaCache` files. I think this might becomes more clear with my next set of changes 😁.

hmartinez82:
@jxy-s I'm closing this. Freel free to take whatever is needed into your working branch.

jxy-s:
Thanks @hmartinez82 I'll credit your work in the commit ❤️

PR #1996, opened 2024-03-10
Service Tag for Informer Messages

DavidXanatos:
This change adds the acting threads service tag to informer messages, granting additional insight whom did what.

### DIFF
--- a/KSystemInformer/util.c
+++ b/KSystemInformer/util.c
@@ -2316,11 +2316,15 @@ UINT_PTR KphGetCurrentThreadSubProcessTag(
     ULONG_PTR SubProcessTag = 0;
 
     __try {
+
         PVOID Teb = PsGetCurrentThreadTeb();
 
-        UINT_PTR* pSubProcessTag = (UINT_PTR*)(((UINT_PTR)Teb) + FIELD_OFFSET(TEB, SubProcessTag));
+        if (Teb) {
 
-        SubProcessTag = *pSubProcessTag;
+            UINT_PTR* pSubProcessTag = (UINT_PTR*)(((UINT_PTR)Teb) + FIELD_OFFSET(TEB, SubProcessTag));
+
+            SubProcessTag = *pSubProcessTag;
+        }
     }
     __except (EXCEPTION_EXECUTE_HANDLER) {
     }
--- a/KSystemInformer/util.c
+++ b/KSystemInformer/util.c
@@ -2301,3 +2301,36 @@ NTSTATUS KphDominationAndPrivilegeCheck(
                               ProcessTarget,
                               AccessMode);
 }
+
+/**
+ * \brief Returns the service tag of the current thread on success and 0 on failure.
+ *
+ * \return Service tag .
+ */
+_IRQL_requires_max_(PASSIVE_LEVEL)
+_Must_inspect_result_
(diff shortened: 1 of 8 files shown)


kphlib/include/kphmsgdefs.h
jxy-s, line 1: Are there other messages that could benefit from this? 

This PR can't be accepted immediately into `master`, changes the offsets of message data requires bumping the `KPH_MESSAGE_VERSION`, else clients will read incorrect message data. I have a staging branch going with changes where it would be most appropriate to incorporate such a change. Would it be okay for me to incorporate this request there instead?

jxy-s (changes requested):

KSystemInformer/util.c
```diff
@@ -2301,3 +2301,36 @@ NTSTATUS KphDominationAndPrivilegeCheck(
...
+        ULONG_PTR Teb;
+#ifdef _M_ARM64
+        Teb = (*((ULONG_PTR*)(__getReg(18) + 0x30)));
+#elif _WIN64
+        Teb = (ULONG_PTR)__readgsqword(0x30);
+#else ! _WIN64
+        Teb = (ULONG_PTR)__readfsdword(0x18);
+#endif _WIN64
+
+        ULONG* pSvcTag = (ULONG*)((PVOID)(Teb + FIELD_OFFSET(TEB, SubProcessTag)));
```
jxy-s, line 2328: `SubProcessTag` is 64 or 32 bits, I think we should capture the information as it's presented for the sake of future-proofing including `SubProcessTag` in messages. 

Might also be worth calling the API (and message members) it what it is "SubProcessTag" rather than "ServiceTag". I recolonize that's what it is used for, but it might be worth just naming it after what is in the `TEB` for the sake of clarity?

kphlib/include/kphmsgdefs.h
DavidXanatos: Sure wherever it fits best.

KSystemInformer/util.c
DavidXanatos: yes makes sense, lets go with a 64bit SubProcessTag

KSystemInformer/util.c
```diff
@@ -2301,3 +2301,36 @@ NTSTATUS KphDominationAndPrivilegeCheck(
...
+
+    __try {
+        ULONG_PTR Teb;
+#ifdef _M_ARM64
+        Teb = (*((ULONG_PTR*)(__getReg(18) + 0x30)));
+#elif _WIN64
+        Teb = (ULONG_PTR)__readgsqword(0x30);
+#else ! _WIN64
+        Teb = (ULONG_PTR)__readfsdword(0x18);
+#endif _WIN64
```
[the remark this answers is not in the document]
DavidXanatos: That's how I always did it but indeed this function is apparently available since windows 7 so we can use it :D

kphlib/include/kphmsgdefs.h
jxy-s: Okay, I'll cherry-pick your commit into the branch and get it merged with the other changes 👍 - we can close this PR then.

DavidXanatos: ok shell I close it now or once the changes are picked?

Issue #1083, opened 2022-01-01
Additional metadata for processors on CPU details (graphs) display?


robertroessler:
As we begin to explore the new worlds of Windows 11 and 12-th Gen Intel CPUs - and their interactions - it could be nice to be able to see some additional data for each of the "logical processors".

Basically, I have 3 possible "enhancements" so far:

1) for Win10/Win11, some visual indication of the physical cores _and_ their associated logical processors (like a very thin dotted line around each physical core's grouping? Yes, this is a very preliminary suggestion, and doesn't handle very well logical processors that are on the same core but on different rows of the display)

2) an indication of the type of physical core "type" (like **P** or **E** at the start of the %-usage display?)

3) an indication of "parked" status (like replacing the graph area of a parked core / logical processor with cross-hatching?) - and yes, this may have relevance pre-Windows 11 / 12th-Gen Intel CPUs

Aside from the obvious eye-candy aspects - which may or may not be enough to justify dev time - I am thinking features like these could be of real help in seeing the results of the schedulers (both H/W and S/W) and their interactions with workloads.

And increased visibility into these underlying processes _could_ eventually lead to increased understanding.

dmex:
> indication of the type of physical core "type" (like P or E

The latest nightly build now shows E-Core or P-Core in the tooltip when hovering the mouse over the processor graphs 👍

PR #2040, opened 2024-04-23
Display core parking indicator when hovering the mouse over the processor graphs

wnstngs:
## Purpose
Starting with Windows 7, the thread scheduler and processor power manager are able to park processors for power saving purposes.  Parking a processor makes it unavailable for thread scheduling, which contributes to potentially putting this processor into a deep idle state and thus reducing power consumption. Parked cores are marked as "Parked" in Task Manager, but not in System Informer.

This PR adds support for displaying core parking status in the tooltip when hovering over the processor graphs. Previously there was a proposal to add such a feature: https://github.com/winsiderss/systeminformer/issues/1083

## Proposed changes
1) Function that determines whether a core is parked. This uses the powerful CPU Sets mechanism available since Windows 10. Currently, every logical processor has a CPU Set and this looks like the most convenient and reliable way to determine if a specific core is parked.
2) Display "Parked" in the tooltip if the core is parked. https://imgur.com/a/Z1iDT8W

## TODO?
1) Revise the tooltip formatting code, as in the current form it easily becomes even more confusing when adding new functionality (`format` indexes change depending on different conditions). For example, at least use named variables for indexes of certain components of the tooltip (load percentages, relationships, time, and so on).
2) Consider using the CPU Sets mechanism in more places.


wnstngs:

SystemInformer/syssccpu.c
```diff
@@ -1903,3 +1925,87 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+    cpuSets = PhAllocate(returnedLength);
+    if (cpuSets == NULL)
+    {
+        return FALSE;
+    }
+
+    //
+    // Get system CPU Set information.
+    //
+    if (!GetSystemCpuSetInformation(cpuSets,
```
jxy-s, line 1984: Please use native functions where possible:
`GetCurrentProcess` -> `NtCurrentProcess`
`GetSystemCpuSetInformation` -> `NtQuerySystemInformationEx` + `SystemCpuSetInformation`

SystemInformer/syssccpu.c
```diff
@@ -1078,17 +1079,38 @@ VOID PhSipNotifyCpuGraph(
...
+                            }
+                            
                         }
                         else
                         {
                             PhInitFormatS(&format[13], L"\n");
-                            PhInitFormatSR(&format[14], PH_AUTO_T(PH_STRING, PhGetStatisticsTimeString(NULL, getTooltipText->Index))->sr);
 
-                            PhMoveReference(&CpusGraphState[Index].TooltipText, PhFormat(format, 15, 0));
+                            if (PhIsCoreParked(Index))
```
jxy-s, line 1098: Tooltips can be fussy. Might need to try to cache this information. Might be an opportunity to eliminate an allocation. Agreed the formatting indexes could do some cleanup.

SystemInformer/syssccpu.c
```diff
@@ -1903,3 +1925,87 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+                                    currentProcess,
+                                    0))
+    {
+        goto CleanupExit;
+    }
+
+    //
+    // Check if the set of the target logical processor is parked.
+    //
+    if (cpuSets[ProcessorIndex].CpuSet.Parked) {
```
jxy-s, line 1996: `SYSTEM_CPU_SET_INFORMATION` is documented as variable-sized:
>This is a variable-sized structure designed for future expansion. When iterating over this structure, use the size field to determine the offset to the next structure.

SystemInformer/syssccpu.c
```diff
@@ -1903,3 +1925,87 @@ PPH_STRINGREF PhGetHybridProcessorType(
...
+ * \return TRUE if the specified logical processor is parked, FALSE otherwise or
+ * in case of failure.
+ */
+BOOLEAN PhIsCoreParked(
+    _In_ ULONG ProcessorIndex
+    )
+{
+    PSYSTEM_CPU_SET_INFORMATION cpuSets = NULL;
+    ULONG returnedLength;
+    HANDLE currentProcess = GetCurrentProcess();
```
jxy-s, line 1947: Current process is optional if we aren't interested in `AllocatedToTargetProcess`. Since we aren't interested in that field we should probably pass `NULL` instead?

jxy-s (commented):
@wnstngs thanks for the contribution! I agree with showing the parked state in the tooltip. I have some feedback. Either I can merge this as and and go back to address it myself, or you can do so. Let me know how you'd like to handle it.

SystemInformer/syssccpu.c
wnstngs: Fixed: https://github.com/winsiderss/systeminformer/pull/2040/commits/4458096161c84278c458cbac55f57d1f73b636c1

Now instead of `GetSystemCpuSetInformation` we use `NtQuerySystemInformationEx` which supports the `SystemCpuSetInformation` class.

SystemInformer/syssccpu.c
wnstngs: Nice observation, thanks!
Fixed: https://github.com/winsiderss/systeminformer/pull/2040/commits/5e510ad5bd74836c340c79b0bddf949f550ad1e6

SystemInformer/syssccpu.c
wnstngs: That's right. As of https://github.com/winsiderss/systeminformer/pull/2040/commits/4458096161c84278c458cbac55f57d1f73b636c1 we use `NtQuerySystemInformationEx` in which we don't pass a handle to the actual process (but we do pass a valid indirected handle as requested by the routine).

SystemInformer/syssccpu.c
wnstngs: Unlike the other parameters, the parking state of a set can be changed quite often by the processor power manager, so I'm not sure about caching. But I think it'd be worth at least adding a delay of half a second or a bit more.

jxy-s (approved):
Looks reasonable to me as it is, I will merge this when I have time - there is some cleanup I want to do after it gets merged, specifically around the juggling the formatting of the string mentioned in the PR.

dmex (approved):

wnstngs:
@jxy-s Thank you!

PR #2050, opened 2024-05-04
calling convention fixes for better consistency

DavidXanatos:
The changes helps to compile the phlib into projects set up for different default calling conventions (like Task Explorer)

### DIFF
--- a/kphlib/include/kphdyndata.h
+++ b/kphlib/include/kphdyndata.h
@@ -14,7 +14,7 @@
 #include <kphdyn.h>
 
 _Must_inspect_result_
-NTSTATUS KphDynDataGetConfiguration(
+NTSTATUS NTAPI KphDynDataGetConfiguration(
     _In_ PKPH_DYNDATA DynData,
     _In_ ULONG DynDataLength,
     _In_ USHORT MajorVersion,
--- a/phlib/include/apiimport.h
+++ b/phlib/include/apiimport.h
@@ -229,7 +229,7 @@ typedef VOID (WINAPI* _UnsubscribeServiceChangeNotifications)(
     _In_ PSC_NOTIFICATION_REGISTRATION pSubscription
     );
 
-#define PH_DECLARE_IMPORT(Name) _##Name Name##_Import(VOID)
+#define PH_DECLARE_IMPORT(Name) _##Name NTAPI Name##_Import(VOID)
 
 PH_DECLARE_IMPORT(NtQueryInformationEnlistment);
 PH_DECLARE_IMPORT(NtQueryInformationResourceManager);
--- a/phlib/include/appresolver.h
+++ b/phlib/include/appresolver.h
@@ -14,23 +14,23 @@
 
 EXTERN_C_START
 
-HRESULT PhAppResolverGetAppIdForProcess(
+HRESULT NTAPI PhAppResolverGetAppIdForProcess(
     _In_ HANDLE ProcessId,
     _Out_ PPH_STRING *ApplicationUserModelId
     );
 
-HRESULT PhAppResolverGetAppIdForWindow(
+HRESULT NTAPI PhAppResolverGetAppIdForWindow(
     _In_ HWND WindowHandle,
     _Out_ PPH_STRING *ApplicationUserModelId
     );
 
-HRESULT PhAppResolverActivateAppId(
+HRESULT NTAPI PhAppResolverActivateAppId(
     _In_ PPH_STRING ApplicationUserModelId,
     _In_opt_ PWSTR CommandLine,
     _Out_opt_ HANDLE *ProcessId
     );
 
-HRESULT PhAppResolverPackageTerminateProcess(
+HRESULT NTAPI PhAppResolverPackageTerminateProcess(
     _In_ PPH_STRING PackageFullName
     );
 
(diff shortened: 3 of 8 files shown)


kphlib/include/kphdyndata.h
```diff
@@ -14,7 +14,7 @@
 #include <kphdyn.h>
 
 _Must_inspect_result_
-NTSTATUS KphDynDataGetConfiguration(
+NTSTATUS NTAPI KphDynDataGetConfiguration(
```
dmex, line 17: None of the Kph functions are exported or external.

phlib/include/apiimport.h
```diff
@@ -229,7 +229,7 @@ typedef VOID (WINAPI* _UnsubscribeServiceChangeNotifications)(
     _In_ PSC_NOTIFICATION_REGISTRATION pSubscription
     );
 
-#define PH_DECLARE_IMPORT(Name) _##Name Name##_Import(VOID)
+#define PH_DECLARE_IMPORT(Name) _##Name NTAPI Name##_Import(VOID)
```
dmex, line 232: These types don't have calling conventions.

phlib/include/appresolver.h
```diff
@@ -14,23 +14,23 @@
 
 EXTERN_C_START
 
-HRESULT PhAppResolverGetAppIdForProcess(
+HRESULT NTAPI PhAppResolverGetAppIdForProcess(
```
dmex, line 17: The AppResolver functions are not external or exported.

phlib/include/kphcomms.h
```diff
@@ -14,7 +14,7 @@
 
 #include <kphmsg.h>
 
-NTSTATUS KphFilterLoadUnload(
+NTSTATUS NTAPI KphFilterLoadUnload(
```
dmex, line 17: These functions are not exported or external.

phlib/include/phutil.h
```diff
@@ -1853,6 +1853,7 @@ PhHungWindowFromGhostWindow(
 
 PHLIBAPI
 NTSTATUS
+NTAPI
```
dmex, line 1856: ✔️

phlib/include/settings.h
```diff
@@ -376,7 +377,9 @@ VOID PhResetSettings(
 // begin_phapppub
 // High-level settings creation
 
-VOID PhAddSetting(
+VOID
+NTAPI
+PhAddSetting(
```
dmex, line 382: ✔️

dmex (changes requested):

kphlib/include/kphdyndata.h
jxy-s: Agreed, should not be necessary.

DavidXanatos: 3rd party projects which use the kphlib may link against un exported functions, and if they by default use a different calling convention then __stdcall a fix like this is required.

phlib/include/appresolver.h
DavidXanatos: 3rd party projects which use the kphlib may link against un exported functions, and if they by default use a different calling convention then __stdcall a fix like this is required.

phlib/include/kphcomms.h
DavidXanatos: 3rd party projects which use the kphlib may link against un exported functions, and if they by default use a different calling convention then __stdcall a fix like this is required.

phlib/include/apiimport.h
DavidXanatos: well my VS2022 begs to differ without that fix i can not build a x86 version of Task Explorer

jxy-s (changes requested):
Many of the calling convention changes aren't necessary upstream. The request is to support oddity calling conventions that might be default in a secondary project. These are internal libraries and not intended to be exported. If there is a use case of a secondary project, those should be maintained in an alternate fork that is specific to the oddities of that other project.

DavidXanatos:
> Many of the calling convention changes aren't necessary upstream. The request is to support oddity calling conventions that might be default in a secondary project. These are internal libraries and not intended to be exported. If there is a use case of a secondary project, those should be maintained in an alternate fork that is specific to the oddities of that other project.

The calling convention use by VS by default is __cdecl so if anything the default use of __stdcall is an oddity.
Since SI is open source IMHO it is to be expected that other projects will use parts of it especially if they come prepackaged as convenient libraries.
For SI it is a change without a difference but for others its a great help hence I assumed it would be a good idea to be more explicit about the used calling convention.
Its disappointing that such a improvement is not welcome.

jxy-s:
> > Many of the calling convention changes aren't necessary upstream. The request is to support oddity calling conventions that might be default in a secondary project. These are internal libraries and not intended to be exported. If there is a use case of a secondary project, those should be maintained in an alternate fork that is specific to the oddities of that other project.
> 
> The calling convention use by VS by default is __cdecl so if anything the default use of __stdcall is an oddity.
> Since SI is open source IMHO it is to be expected that other projects will use parts of it especially if they come prepackaged as convenient libraries.
> For SI it is a change without a difference but for others its a great help hence I assumed it would be a good idea to be more explicit about the used calling convention.
> Its disappointing that such a improvement is not welcome.

I'm not sure that's accurate. For kphlib, It's an x64 or ARM64 static library linked and compiled into the target binary. So the calling convention shouldn't be `__cdecl`. If it is, we should open an alternate discussion on correcting that problem separately from specifying it explicitly. The project might be misconfigured. But none of those changes would be in support of consuming that from a 3rd party.

DavidXanatos:
> 
> I'm not sure that's accurate. For kphlib, it's an x64 or ARM64 static library linked and compiled into the target binary. So the calling convention shouldn't be `__cdecl`. If it is, we should open an alternate discussion on correcting that problem separately from specifying it explicitly. The project might be misconfigured. But none of those changes would be in support of consuming that from a 3rd party.

x64 and ARM64 each only supports one calling convention which booth somewhat resembles __vectorcall the explicit  calling convention specifier have only a meaning on x86, so the issue only arises when trying to compile a project using kphlib for x86, the others x64 and ARM64 work just fine no matter what.

Process Hacker opted to, for whatever reason, to use __stdcall in its projects, which as described is not the VS default nowadays new projects start out with __cdecl preset.
So any 3rd party project which want to use kphlib needs to workaround this mismatch.

If you don't want to support the use case of kphlib being linked by 3rd parties than this pull request is irrelevant and should be closed.

jxy-s:
> > I'm not sure that's accurate. For kphlib, it's an x64 or ARM64 static library linked and compiled into the target binary. So the calling convention shouldn't be `__cdecl`. If it is, we should open an alternate discussion on correcting that problem separately from specifying it explicitly. The project might be misconfigured. But none of those changes would be in support of consuming that from a 3rd party.
> 
> x64 and ARM64 each only supports one calling convention which booth somewhat resembles __vectorcall the explicit calling convention specifier have only a meaning on x86, so the issue only arises when trying to compile a project using kphlib for x86, the others x64 and ARM64 work just fine no matter what.
> 
> Process Hacker opted to, for whatever reason, to use __stdcall in its projects, which as described is not the VS default nowadays new projects start out with __cdecl preset. So any 3rd party project which want to use kphlib needs to workaround this mismatch.
> 
> If you don't want to support the use case of kphlib being linked by 3rd parties than this pull request is irrelevant and should be closed.
> 
> PS: IMHO fostering 3rd party usage would help to engage more contributions to the project and be in the projects very own self interest.

The driver doesn't support 32-bit. We should probably stop referencing `kphlib` for that build and `ifdef` out all the client kph logic instead.

DavidXanatos:
> The driver doesn't support 32-bit. We should probably stop referencing `kphlib` for that build and `ifdef` out all the client kph logic instead.

Only one of the changes addresses the kphlib all other changes address the user mode phlib.
If we could ensure phlib can be used by projects which don't use __stdcall as default that would already help a great lot :)

jxy-s:
There has been no progress and the recent rewrite caused many conflicts. Closing.

PR #2079, opened 2024-05-29
replace all NtReadVirtualMemory occurrances with safe variant

jdpatdiscord:
Hi, I've ran into issues with a new and innovative anticheat method where an anticheat will posit pages that will never be read by the game and therefore purposefully invalidate the page, and if a function like `NtReadVirtualMemory` is called it causes the page to be marked as valid, which the anticheat detects by doing `QueryWorkingSetEx` or `NtQueryVirtualMemory` otherwise.

The new function I've created has a new setting that is off by default because I have found in testing that Windows naturally invalidates pages and causes random regions of memory to not be read. However it seems to have not caused problems within core functionality thus far and the user just notices missing pages scattered about.

I've marked this PR as a draft for feedback. I am considering adding a UI element to toggle this setting so users can be safe or unsafe easily.

### DIFF
--- a/SystemInformer/anawait.c
+++ b/SystemInformer/anawait.c
@@ -303,7 +303,7 @@ BOOLEAN NTAPI PhpWalkThreadStackAnalyzeCallback(
         PVOID timeoutAddress = StackFrame->Params[1];
         LARGE_INTEGER timeout;
 
-        if (NT_SUCCESS(NtReadVirtualMemory(
+        if (NT_SUCCESS(PhSafeReadVirtualMemory(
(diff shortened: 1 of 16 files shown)


jdpatdiscord:

jxy-s:
Please create a new pull request if this is something you want to pursue. However, I personally do not believe this is something we should accept upstream and that you should maintain on your own fork if that's something you want. But, I recognize it would be nice to have to not trigger sentinel pages when searching memory. That said, I must be clear that defense evasion is not something I can condone.

PR #2083, opened 2024-06-03
Updated ProcessRedirectionTrustPolicy to handle both enforce and audit policies correctly

cssaheel:
Fixed: if both EnforceRedirectionTrust and AuditRedirectionTrust are set, the current implementation would cause the audit portion of the code to overwrite the ShortDescription and LongDescription set by the enforce portion. This would result in losing the enforce description. To handle this situation properly, you can combine the descriptions when both policies are set.


cssaheel:

dmex (approved):

jxy-s (approved):

PR #2105, opened 2024-06-20
Update the prototype of NtCreateIRTimer

elnx:
According to my reverse engineering results, the pseudocode 
 of NtCreateIRTimer syscall is as follows:

```c
NTSTATUS NtCreateIRTimer(PHANDLE TimerHandle, LPVOID Reserved, ACCESS_MASK DesiredAccess) {
  return NtCreateTimer2(TimerHandle, Reserved, 0, 2, DesiredAccess);
}
```


dmex:
What parameters does this kernel have for NtCreateIRTimer?

[ntoskrnl.zip](https://github.com/user-attachments/files/15934613/ntoskrnl.zip)

image

elnx:
It looks like this syscall has a different prototype on different version? I reserved this syscall on a win10 ntoskrnl. Maybe a version determination macro is needed here.

phnt/include/ntexapi.h
```diff
@@ -721,6 +721,7 @@ NTSTATUS
 NTAPI
 NtCreateIRTimer(
     _Out_ PHANDLE TimerHandle,
+    _In_ LPVOID Reserved,
```
dmex, line 724: * LPVOID needs to be PVOID
* `PVOID Reserved` would be `HANDLE IRTimerHandle` on Windows 10 and 11?

dmex (changes requested):

dmex (approved):

PR #2121, opened 2024-07-01
Fix environment variable parsing error for CustomBuildTool

pedoc:
For example, environment variables such as ``ESET_HOME=`` will cause the parameter range to overflow.


pedoc:

dmex (approved):

PR #2125, opened 2024-07-02
Remove duplicate definitions in phnt

m417z:


phnt/include/ntpsapi.h
```diff
@@ -684,9 +684,6 @@ typedef struct _PROCESS_MITIGATION_REDIRECTION_TRUST_POLICY
 #endif
 
 #if !defined(NTDDI_WIN10_NI) || (NTDDI_VERSION < NTDDI_WIN10_NI)
-#define ProcessUserPointerAuthPolicy 17
-#define ProcessSEHOPPolicy 18
```
dmex, line 688: These two are version checked and required for older SDK's

dmex (changes requested):
```
#define ProcessUserPointerAuthPolicy 17
#define ProcessSEHOPPolicy 18
```

phnt/include/ntpsapi.h
m417z: They're defined unconditionally below:
https://github.com/winsiderss/systeminformer/blob/8f05a6009e7b04dc4370b1ad2b10d46183db186a/phnt/include/ntpsapi.h#L742-L745

What am I missing?

dmex (approved):

PR #2144, opened 2024-07-30
Cleanup process state and power request objects on process exit

diversenok:
If the user freezes of requests execution for a process and then the process terminates, System Informer doesn't free up the process state and the power requests object. This pull requests fixes this issue.


dmex (commented):
The delete procedure is asynchronous and we can't use locks here.  I'll have to rewrite these and remove the hashtables.

PR #2148, opened 2024-08-07
Fixed set handle attributes feature

DartVanya:
Tested on Win11 23H2

https://github.com/user-attachments/assets/8f0d009a-fc0b-43ee-8f67-400c31c147e0

### DIFF
--- a/SystemInformer/prpghndl.c
+++ b/SystemInformer/prpghndl.c
@@ -81,7 +81,7 @@ VOID PhpInitializeHandleMenu(
     _Inout_ PPH_HANDLES_CONTEXT HandlesContext
     )
 {
-    PPH_EMENU_ITEM item;
+    //PPH_EMENU_ITEM item;
     KPH_LEVEL level;
 
     if (NumberOfHandles == 0)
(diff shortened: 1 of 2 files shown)


DartVanya:

SystemInformer/prpghndl.c
```diff
@@ -110,13 +110,13 @@ VOID PhpInitializeHandleMenu(
...
     level = KsiLevel();
 
-    if (level >= KphLevelMed)
-    {
-        if (item = PhFindEMenuItem(Menu, 0, NULL, ID_HANDLE_PROTECTED))
-            PhDestroyEMenuItem(item);
-        if (item = PhFindEMenuItem(Menu, 0, NULL, ID_HANDLE_INHERIT))
-            PhDestroyEMenuItem(item);
-    }
+    //if (level >= KphLevelMed)
```
jxy-s, line 113: Looking back at history this check should have been `if (level < KphLevelMed)` instead of `>=`, oops. Nice find!

I think we should just deleted the commented out code entirely.

If we present users with the options and they use them with the driver disabled, a message will inform the user to enable the driver. An alternative would be to disable the menu item `PH_EMENU_DISABLED` when `if (level < KphLevelMed)`, but I think maybe the best experience is to just leave those options there and enabled.

jxy-s (changes requested):

SystemInformer/prpghndl.c
DartVanya: Ok, I removed commented lines entirely

jxy-s: Thanks for doing that. And yep that is the intended required access for this path 👍.

jxy-s (approved):

PR #2193, opened 2024-09-11
Buildtools signing

dmex:


ge0rdi:
@dmex
Wondering when this PR could be merged.

It contains fix for SourceLink (among other things), that would be nice to have working.
Maybe you can merge at least that fix? So that future PDBs will contain proper SourceLink information.

jxy-s:
> @dmex Wondering when this PR could be merged.
> 
> It contains fix for SourceLink (among other things), that would be nice to have working. Maybe you can merge at least that fix? So that future PDBs will contain proper SourceLink information.

In this this branch needs more testing? If we want the SourceLink works we should just pull that forward on its own.

ge0rdi:
> If we want the SourceLink work we should just pull that forward on its own.

That would be really cool.
I have created fix some time ago (#2203), but it was closed because this PR contains the fix as well.

PR #2203, opened 2024-09-15
Fix SourceLink

ge0rdi:
`SourceLink` used to work at some point, but doesn't work with recent SI builds.

The reason is that SourceLink stream in PDBs isn't valid json anymore.
It contains _doubled_ `{{` / `}}` instead of `{` / `}`.
Also there shouldn't be coma after the last item in dictionary.


dmex:
This already has a PR: https://github.com/winsiderss/systeminformer/pull/2193

ge0rdi:
@dmex 
Cool, didn't notice that.
Hopefully that PR will get merged soon.

PR #2208, opened 2024-09-17
Fix EnableStartAsAdminAlwaysOnTop feature

DartVanya:
Previously:
`status = PhGetWindowClientId(GetDesktopWindow(), &desktopId);`
returns csrss PID, and service failed when trying to open its token, because of PPL.
We need to use current SI token, and not linked, sice we already an admin when we try to restart with UiAccess.


dmex:
> We need to use current SI token, and not linked, sice we already an admin 

This code originally used the current token but it prevents the application from running when elevated by UAC from a different session. I'll revert some of the changes.

PR #2222, opened 2024-10-01
Fix RtlGet(System/User)PreferredUILanguages SAL

RatinCN:
The second parameter of `RtlGet(System/User)PreferredUILanguages` can be `NULL`, and this is what `Get(System/User)PreferredUILanguages` did on the latest Win11.


RatinCN:

dmex (approved):

PR #2228, opened 2024-10-06
Master

DartVanya:
Working with PPL process I accidentally noticed that some of them can be terminated. But I was very confused why PH with driver on cannot terminate process which Task Manager can terminate easily.
Some PPL levels designed to allow termination (apparently, I didn't even know that).
This levels can be terminated:
- 0\. PsProtectedSignerNone
- 1\. PsProtectedSignerAuthenticode
- 2\. PsProtectedSignerCodeGen
- 5\. PsProtectedSignerWindows,
- 8\. PsProtectedSignerApp.

When driver enabled PhTerminateProcess tries to use KSI, and because of how KSI domination check works it return access denided, despite what process was already oppened with PROCESS_TERMINATE.
I suggest to change status check in this way to allow NtTerminateProcess try to terminate process anyway.

Example without driver (NtTerminateProcess only):
https://github.com/user-attachments/assets/e95a4050-0292-4fbd-81e4-7179acfdecf6

Example with driver:
https://github.com/user-attachments/assets/36c6c9df-45a4-4c27-8991-72d7daa708dc


jxy-s:
Yep, the domination check is rudimentary because of missing exports from Microsoft. We should have been trying `NtTerminateProcess` to begin with. Your PR appears to have a bunch of unnecessary commits? I went ahead and fixed it here: https://github.com/winsiderss/systeminformer/commit/9d8309d82d3cb751e4a26c891ef4e2efc0579081 - Closing this PR 👍

PR #2238, opened 2024-10-13
Actual darkmode fix

Factor101:


dmex:
I'll make these customizable via application settings

PR #2254, opened 2024-10-25
kph Don't map whole section when looking for mapped file name

ge0rdi:
It is enough to map just first page of section in order to query mapped file name.
Mapping whole section may cause unnecessarily big paged pool usage in case of huge sections.

---

The issue with big pool usage can be easily demonstrated this way:

- Open some project in Visual Studio 2022
- Notice current paged pool size (either in SI or Task Manager)
- In SI press Ctrl+F, enter some string and press enter
- Paged pool size raises by ~20GB

Visual Studio has about 5 processes where each has opened handle to 2TB section (`Reserve (2 TB)`). Running search in SI will cause that each such section is mapped (which takes ~4GB of paged pool), but the memory is not freed upon section unmapping. It is freed when given section handle closes (so you need to close Visual Studio in order to reclaim that memory).

This seems to be standard Windows behavior (at least Win10/11). It seems the system caches memory that was allocated for mapping purposes (PTEs) so that it can be immediately reused when section is mapped next time.

### DIFF
--- a/KSystemInformer/object.c
+++ b/KSystemInformer/object.c
@@ -2021,7 +2021,7 @@ NTSTATUS KphQueryInformationObject(
             KeStackAttachProcess(PsInitialSystemProcess, &apcState);
 
             baseAddress = NULL;
-            viewSize = 0;
+            viewSize = PAGE_SIZE;
             status = ZwMapViewOfSection(sectionHandle,
                                         ZwCurrentProcess(),
                                         &baseAddress,
@@ -2056,9 +2056,9 @@ NTSTATUS KphQueryInformationObject(
                 ZwUnmapViewOfSection(ZwCurrentProcess(), baseAddress);
             }
 
-            ObCloseHandle(sectionHandle, KernelMode);
-
             KeUnstackDetachProcess(&apcState);
+
+            ObCloseHandle(sectionHandle, KernelMode);
 
             if (NT_SUCCESS(status))
             {


ge0rdi:
I have also noticed that [ObCloseHandle](https://github.com/winsiderss/systeminformer/blob/master/KSystemInformer/object.c#L2059) should be probably done after [KeUnstackDetachProcess](https://github.com/winsiderss/systeminformer/blob/master/KSystemInformer/object.c#L2061).
Do you think I can fix that as well?

jxy-s:
Nice find @ge0rdi. I think we could do even better. Since we no longer have to support Windows 7 I think there is an ideal way that will almost always reclaim immediately and avoid having to invade the system process. Allow me some time to investigate.

jxy-s (changes requested):
Nice find @ge0rdi. The kernel is going to complete the accounting for the section if we map the entire thing. Which is why the pool usage jump for very large sections.

> I have also noticed that [ObCloseHandle](https://github.com/winsiderss/systeminformer/blob/master/KSystemInformer/object.c#L2059) should be probably done after [KeUnstackDetachProcess](https://github.com/winsiderss/systeminformer/blob/master/KSystemInformer/object.c#L2061).
Do you think I can fix that as well?

This doesn't technically matter. But for correctness of ordering releasing of resources - yes go ahead and re-order those.

Please conform to the commit message style for kernel driver changes:
`kph all lower case subject`

jxy-s (approved):

ge0rdi:
@jxy-s 
Do you have any idea when this could get released?
It is rather annoying issue (though I guess it is much better now that I understand what is causing it).

jxy-s:
> @jxy-s 
> Do you have any idea when this could get released?
> It is rather annoying issue (though I guess it is much better now that I understand what is causing it).

No, but I'll do what I can to expedite it. Maybe a few weeks. Depends on when we can submit for signing.

ge0rdi:
Cool, thank you.

> Depends on when we can submit for signing.

What do you mean by that? I guess you can submit driver for attestation signing at any time.
Or is there something you need to wait for?

jxy-s:
> What do you mean by that? I guess you can submit driver for attestation signing at any time.
Or is there something you need to wait for?

Just people time to do the signing and test the changes before doing a build.

Took care of it already - `3.1.24305.0` has the updated driver 👍

PR #2257, opened 2024-10-27
Few improvements

ge0rdi:
- fix Ctrl+K (search box) handling
  - tree view didn't pass keypresses to parent window so it was not possible to handle the keyboard combo
  - some plugins didn't handle Ctrl+K at all
- treenew handle Ctrl+A in common code
- searchbox pressing escape will abandon search

  Current search is abandoned and focus is returned to window that previously had it, instead of closing whole dialog.

### DIFF
--- a/SystemInformer/searchbox.c
+++ b/SystemInformer/searchbox.c
@@ -53,6 +53,7 @@ typedef struct _PH_SEARCHCONTROL_CONTEXT
     };
 
     HWND ParentWindowHandle;
+    HWND PreviousFocusWindowHandle;
     LONG WindowDpi;
 
     PH_SEARCHCONTROL_BUTTON SearchButton;
@@ -1024,6 +1025,9 @@ LRESULT CALLBACK PhpSearchWndSubclassProc(
     case WM_KILLFOCUS:
         RedrawWindow(hWnd, NULL, NULL, RDW_FRAME | RDW_INVALIDATE);
         break;
+    case WM_SETFOCUS:
+        context->PreviousFocusWindowHandle = (HWND)wParam;
+        break;
     case WM_SETTINGCHANGE:
     case WM_SYSCOLORCHANGE:
     case WM_THEMECHANGED:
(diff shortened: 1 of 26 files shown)


ge0rdi:
`Ctrl+K` shortcut should now work in all dialogs that have search box.

I think the behavior of Esc (while search box has focus) is much better now. Search is cancelled and focus returned to where it was prior to doing search.

SystemInformer/searchbox.c
```diff
@@ -1211,6 +1215,17 @@ LRESULT CALLBACK PhpSearchWndSubclassProc(
                     }
                 }
             }
+            if (wParam == VK_ESCAPE)
```
jxy-s, line 1218:
```suggestion
            else if (wParam == VK_ESCAPE)
```

SystemInformer/searchbox.c
```diff
@@ -1211,6 +1215,17 @@ LRESULT CALLBACK PhpSearchWndSubclassProc(
...
+            {
+                PhSetWindowText(hWnd, L"");
+                PhpSearchUpdateText(hWnd, context, FALSE);
+                if (context->PreviousFocusWindowHandle)
+                {
+                    SetFocus(context->PreviousFocusWindowHandle);
+                    context->PreviousFocusWindowHandle = NULL;
+                }
+                return 1;
+            }
```
jxy-s, line 1228: Would it be possible to make it so up and down arrows also return focus to the previous window (without clearing the text)? Those are common keys to navigate up and down in the tree view but are not generally keys used for navigating the single row search box. That also might provide a nicer user experience to break away from the search box.

phlib/treenew.c
```diff
@@ -1258,6 +1258,18 @@ VOID PhTnpOnKeyDown(
...
         return;
+
+    // handle standard key presses
+    switch (VirtualKey)
+    {
+    case 'A':
+        if (GetKeyState(VK_CONTROL) < 0)
+            TreeNew_SelectRange(hwnd, 0, -1);
+        return;
+    }
```
jxy-s, line 1269: @dmex is this okay to do with treenew? Any reason not to do this?

jxy-s (changes requested):

SystemInformer/searchbox.c
ge0rdi: Good idea, done. 
Seem to work nicely.

phlib/treenew.c
ge0rdi: I think it is common (expected) behavior of tree/list views to be able to select all items.

Most of them had explicit handling of `Ctrl+A`. Though, some didn't (maybe just by mistake?).

It will be still possible to _disable_ `Ctrl+A` for particular tree view (if not wanted) by explicit handling the event and marking it as handled (`keyEvent.Handled = true`).

jxy-s (approved):

Issue #2074, opened 2024-05-24
Show "??" entry in Object Manager


alabuzhev:
### Description of the feature, modification, idea or suggestion

Tools - System - Object Manager shows the content of "\\" by calling `NtQueryDirectoryObject` recursively.
This approach overlooks one particular directory - `??` (a.k.a. "local `\DosDevices`"), which is not returned when querying the root directory, unlike the global version of `\DosDevices` (`GLOBAL??`).
This directory typically contains a symbolic link to `Global`, network drives, subst drives and whatever else the user defined with `DefineDosDevice`.

This information is also available under `\Sessions\<SESSION_ID>\DosDevices\<LOGON_SESSION_ID>`, but getting there is more cumbersome and requires Administrator rights.

### Proposed implementation details (optional)

It can be added manually (I'm not familiar with the code, so apologies if something is horribly wrong):
```DIFF
diff --git a/plugins/ExtendedTools/objmgr.c b/plugins/ExtendedTools/objmgr.c
index 229e7b279..efb1b83d6 100644
--- a/plugins/ExtendedTools/objmgr.c
+++ b/plugins/ExtendedTools/objmgr.c
@@ -14,6 +14,8 @@
 #include <hndlinfo.h>
 
 static PH_STRINGREF EtObjectManagerRootDirectoryObject = PH_STRINGREF_INIT(L"\\"); // RtlNtPathSeperatorString
+static PH_STRINGREF EtObjectManagerUserDirectoryObject = PH_STRINGREF_INIT(L"??");
+static PH_STRINGREF DirectoryObjectType = PH_STRINGREF_INIT(L"Directory");
 static HWND EtObjectManagerDialogHandle = NULL;
 static HANDLE EtObjectManagerDialogThreadHandle = NULL;
 static PH_EVENT EtObjectManagerDialogInitializedEvent = PH_EVENT_INIT;
@@ -875,6 +877,14 @@ INT_PTR CALLBACK WinObjDlgProc(
                 EtObjectManagerRootDirectoryObject
                 );
 
+            DIRECTORY_ENUM_CONTEXT enumContext;
+
+            enumContext.TreeViewHandle = context->TreeViewHandle;
+            enumContext.RootTreeItem = context->RootTreeObject;
+            enumContext.DirectoryPath = EtObjectManagerRootDirectoryObject;
+
[...]

PR #2284, opened 2024-11-17
PhReInitializeTheme: runtime theme switching on ImmersiveColorSet message

DartVanya:
### General
- Theme can be changed at `runtime w/o restart` by `EnableThemeSupport` and `EnableThemeUseWindowsTheme`. All PH/PV windows and dialogs will be redrawed automatically.
- If `EnableThemeUseWindowsTheme` enabled theme will be `switched automatically` when Windows theme is changing.
- Streamer mode, acrylic, theme custom colors settings are applied immediately on all windows w/o restart.
- New option `EnableThemeUseWindowsTheme` - `Use Windows app color mode`. On default settings PH will use current Windows theme.

---
### phlib and other UI fixes
- Moved delayhook to phlib.
- phlib `PhIsThemeSupportEnabled`, `PhReInitializeTheme`, `PhReInitializeStreamerMode`. Exported for plugins.
- New option `EnableThemeTabBorders` (default 0).
- TreeNew: fix custom `ForeColor` for dark mode.

---
### SDK
- `hndlprp.h`. Removed duplicate declarations for plugins. Consolidated handle property APIs in public part of header.
- Add some missing declarations and exports of `IListView` wrappers.
- Update `objmgr` and `objprp` code to latest SDK.
---

> If AppMode is set to PreferredAppModeDarkOnDark, PhShouldAppsUseDarkMode should return current Windows app color mode.
> It's weird that sometimes it can return incorrect inverted value (idk why).
> In this case use PhGetAppsUseLightTheme() to retrieve raw value from registry.
> 
> When app mode set to always dark/light all subsequent calls to PhShouldAppsUseDarkMode always ignore
> the current Windows app color mode. Possible return values are:

| PreferredAppMode | PhShouldAppsUseDarkMode return  | PhGetAppsUseLightTheme return
| ------------- |:-------------:|:-------------:|
| PreferredAppModeDisabled           | current Windows app color mode  | current Windows app color mode |
| PreferredAppModeDarkOnDark     | current Windows app color mode   | current Windows app color mode |
| PreferredAppModeDarkAlways      | always TRUE     | current Windows app color mode |
| PreferredAppModeLightAlways      | always FALSE   | current Windows app color mode | 

https://github.com/user-attachments/assets/cbe024b8-4c02-41bb-85f7-a07ceed5d051

PR #2294, opened 2024-11-23
objmgr cleanup and improvements

DartVanya:
### Object Manager
- [objmgr cleanup and fixes](https://github.com/winsiderss/systeminformer/pull/2284/commits/f754e7b6135748c70128e01c9b82f87f9705e342)
   - `KphObjectAttributesInformation -> KphObjectExtendedInformation`, retrieve object address, type index and flags.
   - Public hndlmenu.h APIs. `ID_HANDLE_OBJECTPROPERTIES[1/2]` menu support for Device/File, Key, Section handles. Add Security menu item.
   - Add Name and Original name columns for Device/File, Key, ALPC Port and for any type object from \ObjectTypes.

- [objmgr cleanup and improvements](https://github.com/winsiderss/systeminformer/pull/2284/commits/bdc1cdf5ee9a8ecfcacfd1d4f6c79132de82d87d)
   - Handles tab Refresh (F5) button added. Handles and refs counts on main page will also be updated.
   - EtTreeViewFindItem add reverse scan for more reliable result.
   - [Resolver] Extract Section information for more kernel objects using driver.
- Add menu - `Copy Object Address (Ctrl+Shift+C)`. Add `Ctrl+Alt+C` shortcut for `Copy Full Name`.
- Show more correct Reference count in object properties (subtracting RefCnt of PH's opened handle). This is light version of windbg !trueref. For permanent objects with no handles it show exact RealPointerCount.
- [Add "Object address" column (showed only when driver on)](https://github.com/winsiderss/systeminformer/pull/2294/commits/172199209a86265b71a5d46ab28f1567c8dac418)  _But copying address from the menu always works._
- [Menu "Handles" added (quick open of Handles tab)](https://github.com/winsiderss/systeminformer/pull/2294/commits/fb392bffe5111744c8c8b28842545fb91d5925f7)
- [Resolver] Show Type info in target column in format: Index: \<n\>, Objects: \<m\>, Handles: \<k\>.
- `One object - one properties window`. Flash already open window to user.
- _Improved navigation to symlinks._
   - Show volume letter in appropriate targets, for example: \\Device\\HarddiskVolume4 (C:)
      - Menu - "Show drive volume" for this cases.
- fix https://github.com/winsiderss/systeminformer/issues/2074
---
### General
- Handle list - new column `"Handle pointer count"`. Show `nt!_HANDLE_TABLE_ENTRY RefCnt`.
   > [Alex Ionescu] Each time an existing handle to an object is used, the reference count goes down by 1.

   Allows to see how processes uses handles in realtime. Minimal code changes, use `Reserved` filed in `KPH_PROCESS_HANDLE` and `SYSTEM_HANDLE_TABLE_ENTRY_INFO_EX`. Works only then driver on, column will be empty in any other scenarios.
- Handle properties - Add `"Handle pointer count"` field. Dynamically updates on `GeneralCallbackProcessProviderUpdatedEvent`.
- [Fix handle list "File share access"](https://github.com/winsiderss/systeminformer/pull/2294/commits/4c45554f711e45560771e7997a3a43efcdb8878e)
- Handle attributes search support. ColorPartiallySuspended for both Protected and Inherit handles.
---
- memsrcht show memory editor on double click, F5 - refresh.
@jxy-s new memory strings dialog is amazing :rocket:
- Improve PhUiCloseHandles for protected handles.
- Fix options saving if dialog closes, when any non-general page active.

#### Preview:

https://github.com/user-attachments/assets/91aa5fae-e72d-402b-a6b2-b24c3b6d656d

### DIFF
--- a/plugins/ExtendedTools/objmgr.c
+++ b/plugins/ExtendedTools/objmgr.c
@@ -117,6 +117,7 @@ typedef struct _ET_OBJECT_ENTRY
     PPH_STRING TargetDrvLow;
     PPH_STRING TargetDrvUp;
     CLIENT_ID TargetClientId;
+    ULONG TypeTypeIndex;
 
     PVOID Object;
     ULONG Attributes;
@@ -676,6 +677,7 @@ static BOOLEAN NTAPI EtEnumCurrentDirectoryObjectsCallback(
         {
             PPH_STRING fullPath;
             PPH_STRING linkTarget;
+            PPH_STRING driveLetter;
 
             fullPath = PH_AUTO(EtGetObjectFullPath(entry->BaseDirectory, entry->Name));
 
@@ -683,6 +685,14 @@ static BOOLEAN NTAPI EtEnumCurrentDirectoryObjectsCallback(
                 NT_SUCCESS(PhQuerySymbolicLinkObject(&linkTarget, NULL, &fullPath->sr)))
             {
                 PhMoveReference(&entry->Target, linkTarget);
+
+                if (PhStartsWithString2(entry->Target, L"\\Device\\HarddiskVolume", TRUE) &&
+                    (driveLetter = PhResolveDevicePrefix(&entry->Target->sr)))
+                {
+                    entry->TargetDrvLow = PhReferenceObject(entry->Target); // HACK
+                    PhMoveReference(&entry->Target, PhFormatString(L"%s (%s)", PhGetString(entry->Target), PhGetString(driveLetter)));
+                    PhDereferenceObject(driveLetter);
+                } 
             }
         }
         else if (entry->EtObjectType == EtObjectWindowStation)
@@ -701,7 +711,8 @@ static BOOLEAN NTAPI EtEnumCurrentDirectoryObjectsCallback(
             entry->EtObjectType == EtObjectSection ||
             entry->EtObjectType == EtObjectEvent ||
             entry->EtObjectType == EtObjectTimer ||
-            entry->EtObjectType == EtObjectSemaphore
+            entry->EtObjectType == EtObjectSemaphore ||
+            entry->EtObjectType == EtObjectType
             )
         {
             entry->TargetIsResolving = TRUE;
@@ -711,15 +722,6 @@ static BOOLEAN NTAPI EtEnumCurrentDirectoryObjectsCallback(
     }
 
     return TRUE;
-}
-
-NTSTATUS EtTreeViewEnumDirectoryObjects2(
-    _In_ HWND TreeViewHandle,
-    _In_ HTREEITEM RootTreeItem,
-    _In_ PPH_STRINGREF DirectoryPath
-    )
-{
-    return EtTreeViewEnumDirectoryObjects(TreeViewHandle, RootTreeItem, PH_AUTO_T(PH_STRING, PhCreateString2(DirectoryPath)));
 }
 
 NTSTATUS EtTreeViewEnumDirectoryObjects(
@@ -1204,6 +1206,35 @@ NTSTATUS EtpTargetResolverWorkThreadStart(
                 }
             }
             break;
+        case EtObjectType:
+            {
+                POBJECT_TYPES_INFORMATION objectTypes;
+                POBJECT_TYPE_INFORMATION objectType;
+                ULONG typeIndex;
+
+                typeIndex = PhGetObjectTypeNumber(&entry->Name->sr);
+
+                if (typeIndex != ULONG_MAX &&
+                    NT_SUCCESS(PhEnumObjectTypes(&objectTypes)))
+                {
+                    objectType = PH_FIRST_OBJECT_TYPE(objectTypes);
+
+                    for (ULONG i = 0; i < objectTypes->NumberOfTypes; i++)
+                    {
+                        if (objectType->TypeIndex == typeIndex)
+                        {
+                            entry->TypeTypeIndex = typeIndex;
+                            entry->TargetIsInfoOnly = TRUE;
+                            entry->Target = PhFormatString(L"Index: %d, Objects: %d, Handles: %d", objectType->TypeIndex, objectType->TotalNumberOfObjects, objectType->TotalNumberOfHandles);
+
+                            break;
+                        }
+                        objectType = PH_NEXT_OBJECT_TYPE(objectType);
+                    }
+                    PhFree(objectTypes);
+                }
+            }
+            break;
     }
 
     if (KsiLevel() >= KphLevelMed && (entry->EtObjectType != EtObjectAlpcPort || alpcSourceOpened)) // show port address if source object was already opened
@@ -1297,7 +1328,7 @@ NTSTATUS EtEnumCurrentDirectoryObjects(
     PH_SORT_ORDER sortOrder;
     WCHAR string[PH_INT32_STR_LEN_1];
 
-    Context->CurrentPath = EtGetSelectedTreeViewPath(Context);
+    PhMoveReference(&Context->CurrentPath, EtGetSelectedTreeViewPath(Context));
 
     ET_CREATE_DIRECTORY_OBJECT_CONTEXT(Context->CurrentPath, objectContext);
 
@@ -2014,7 +2045,10 @@ INT NTAPI WinObjTargetCompareFunction(
     PET_OBJECT_ENTRY item1 = Item1;
     PET_OBJECT_ENTRY item2 = Item2;
 
-    return PhCompareStringZ(PhGetStringOrEmpty(item1->Target), PhGetStringOrEmpty(item2->Target), TRUE);
+    if (item1->EtObjectType == EtObjectType && item2->EtObjectType == EtObjectType)
+        return uintcmp(item1->TypeTypeIndex, item2->TypeTypeIndex);
+    else
+        return PhCompareStringZ(PhGetStringOrEmpty(item1->Target), PhGetStringOrEmpty(item2->Target), TRUE);
 }
 
 INT NTAPI WinObjTriStateCompareFunction(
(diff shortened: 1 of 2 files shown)


DartVanya:
I think this PR is ready to review. :blush:

KSystemInformer/object.c
jxy-s, line 1: Please remove the changes to implement a decoding the true reference. They should be done independently in a separate PR. I'll take care of it myself.

jxy-s (changes requested):

plugins/ExtendedTools/objmgr.c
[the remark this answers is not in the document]
DartVanya: Yes, I notice, what it will lead to access violation. In first version of resolver (when it was only one thread w/o workqueue) I tried to wait and it ends up with deadlock. But I think now, since we workqueue we can wait for main resolver thread from UI.
Thank for recommendation. I try to improve implementation and commit changes. :thumbsup:

KSystemInformer/object.c
DartVanya: Should I open new PR with this changes?

jxy-s: No, I assigned the issue to myself. I already started the work myself. Just remove the changes from this PR, please.

DartVanya: Thanks. I'll remove KphObjectExtendedInformation too. Is it possible to make API similar? Address is usefull for object manager. This existing API the only good way is enumerate NtCurrentProcess() handles, but we to query only one object, and it little bit overkill.

jxy-s: Yes, I'll look into it 👍

plugins/ExtendedTools/objmgr.c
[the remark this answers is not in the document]
DartVanya: Fixed ce31807283ef479750c5b116c4e6eca8dd180d9b

https://github.com/user-attachments/assets/fe5ecf9b-35bb-419b-be8c-8e7b6f5a4430

plugins/ExtendedTools/objprp.c
```diff
@@ -245,6 +258,37 @@ typedef enum _ET_OBJECT_POOLTYPE {
...
+#if 0 // enable this then new driver API will available
+    PKPH_PROCESS_HANDLE_INFORMATION handles;
+    if (KsiLevel() >= KphLevelMed &&
+        NT_SUCCESS(status = KsiEnumerateProcessHandles(NtCurrentProcess(), &handles)))
+    {
+        for (ULONG i = 0; i < handles->HandleCount; i++)
+        {
+            if (handles->Handles[i].Handle == OwnHandle)
+            {
+                *PointerCount = *PointerCount - handles->Handles[i].HandleRefCnt - 1;
```
jxy-s, line 279: Let's please remove this and other areas you have `#if 0` pending driver changes. What you have here isn't correct. An eventual implementation is probably best optimized by letting the driver calculate the real pointer count itself. In order to calculate the real pointer count you need to enumerate all handle tables across the system, not just the current process. The reason you should probably let the driver do this is it can stop enumeration once it has retrieved the `HandleCount` number of handle table entries.

Additionally... I'm not sure if it's even reasonable to be able to calculate the true reference in all cases. The reason being that the real pointer count is a function of the object header pointer count minus the "inverted" ref counts for each active handle. This includes the handles in the CID table (the processes/threads PID/TID table). The CID table is not something immediately available (problem I'm trying to solve in my private branch).

Some background. When the kernel creates a handle it increments the `_OBJECT_HEADER.PointerCount` by `0x8000` and initializes the `_HANDLE_TABLE_ENTRY.RefCnt`. Then when that handle is "used" (e.g. `ObLookupObjectByHandle`) rather than incrementing the `_OBJECT_HEADER.PointerCount` the system simply needs to decrement the `_HANDLE_TABLE_ENTRY.RefCnt`. It then allows the kernel to decrement the `PointerCount` when it has finished with the object. This is "fine" because the `PointerCount` was originally incremented by `0x8000`. When the handle is then closed - the kernel just subtracts the remaining `RefCnt` from the `PointerCount`.

In order to calculate the real pointer count accurately you must enumerate all handle tables on system... the performance of that could be prohibitive of an eventual implementation.

phlib/native.c
```diff
@@ -6914,6 +6914,11 @@ NTSTATUS PhEnumHandlesGeneric(
                 convertedHandles->Handles[i].CreatorBackTraceIndex = 0;
                 convertedHandles->Handles[i].ObjectTypeIndex = handles->Handles[i].ObjectTypeIndex;
                 convertedHandles->Handles[i].HandleAttributes = handles->Handles[i].HandleAttributes;
+#if 0 // enable this then new driver API will available
```
jxy-s, line 6917: Please remove the dead code.

SystemInformer/findobj.c
```diff
@@ -734,6 +734,7 @@ VOID PhpFindObjectAddResultEntries(
         objectNode->BestObjectName = searchResult->BestObjectName;
         objectNode->HandleInfo = searchResult->Info;
         PhPrintPointer(objectNode->HandleString, searchResult->Handle);
+        objectNode->HandleInfo.Reserved = ULONG_MAX;
```
jxy-s, line 737: Is this necessary to keep?

plugins/ExtendedTools/objmgr.c
```diff
@@ -107,22 +128,21 @@ typedef struct _DIRECTORY_ENUM_CONTEXT
...
 typedef struct _RESOLVER_WORK_THREAD_CONTEXT
 {
-    POBJECT_CONTEXT Context;
-    POBJECT_ENTRY entry;
-    INT ItemIndex;
-    //BOOLEAN SortItems;
-} RESOLVER_WORK_THREAD_CONTEXT, * PRESOLVER_WORK_THREAD_CONTEXT;
+    PET_OBJECT_ENTRY entry;
+    PPH_STRING CurrentPath;
+} RESOLVER_WORK_THREAD_CONTEXT, *PRESOLVER_WORK_THREAD_CONTEXT;
```
jxy-s, line 145: You could probably find an implementation that eliminates these entirely since you have access to the `PET_OBJECT_CONTEXT` through the `PET_OBJECT_ENTRY` now. If you don't do this here I'll probably try to optimize it after the merge. Eliminating these will eliminate allocations when resolving the targets.

plugins/ExtendedTools/objmgr.c
```diff
@@ -91,11 +89,34 @@ typedef struct _ET_OBJECT_CONTEXT
...
+    BOOLEAN TargetIsInfoOnly;
+    PPH_STRING TargetDrvLow;
+    PPH_STRING TargetDrvUp;
+    CLIENT_ID TargetClientId;
+
+    PVOID Object;
+    WCHAR ObjectString[PH_PTR_STR_LEN_1];
+} ET_OBJECT_ENTRY, *PET_OBJECT_ENTRY;
 
 typedef struct _DIRECTORY_ENUM_CONTEXT
```
jxy-s, line 121: nit: `ET_DIRECTORY_ENUM_CONTEXT`

plugins/ExtendedTools/objmgr.c
```diff
@@ -107,22 +128,21 @@ typedef struct _DIRECTORY_ENUM_CONTEXT
 typedef struct _HANDLE_OPEN_CONTEXT
```
jxy-s, line 128: nit: `ET_HANDLE_OPEN_CONTEXT`

plugins/ExtendedTools/objmgr.c
```diff
@@ -1158,64 +1300,65 @@ NTSTATUS EtEnumCurrentDirectoryObjects(
...
 
     if (Context->BreakResolverThread)
     {
-        *Context->BreakResolverThread = TRUE;
+        WriteRelease8(Context->BreakResolverThread, TRUE);
         Context->BreakResolverThread = NULL;
     }
 
-    PhClearReference(&Context->CurrentPath);
+    PhClearReference((PVOID)&Context->CurrentPath);
```
jxy-s, line 1322: Cast isn't necessary?

plugins/ExtendedTools/objmgr.c
```diff
@@ -1230,12 +1373,15 @@ NTSTATUS EtDuplicateHandleFromProcessEx(
             0,
             0
             );
-        NtClose(processHandle);
+        if (!ProcessHandle) NtClose(processHandle);
```
jxy-s, line 1376: Will this not leak a handle?

plugins/ExtendedTools/objmgr.c
```diff
@@ -1247,7 +1393,7 @@ NTSTATUS EtDuplicateHandleFromProcessEx(
                 DesiredAccess,
                 Handle
                 );
-            NtClose(processHandle);
+            if (!ProcessHandle) NtClose(processHandle);
```
jxy-s, line 1396: Leaking handle?

plugins/ExtendedTools/objmgr.c
```diff
@@ -1158,64 +1300,65 @@ NTSTATUS EtEnumCurrentDirectoryObjects(
...
-    _In_ POBJECT_CONTEXT Context
+    _In_ PET_OBJECT_CONTEXT Context
     )
 {
     INT index = INT_ERROR;
 
     if (Context->BreakResolverThread)
     {
-        *Context->BreakResolverThread = TRUE;
+        WriteRelease8(Context->BreakResolverThread, TRUE);
```
jxy-s, line 1318: 🤔 - this reads dangerous to me, I think I can reason this away by saying a new thread will not be created and race here? I suppose this is fine then... my initial read is that we could race here and set the wrong boolean context. Or worse a double-fetch of the value resulting in a null pointer de-reference.

plugins/ExtendedTools/objmgr.c
```diff
@@ -1158,64 +1300,65 @@ NTSTATUS EtEnumCurrentDirectoryObjects(
...
-            PhClearReference(&entry->Name);
-            PhClearReference(&entry->TypeName);
-            if (entry->Target)
-                PhClearReference(&entry->Target);
-            if (entry->TargetDrvLow)
-                PhClearReference(&entry->TargetDrvLow);
-            if (entry->TargetDrvUp)
-                PhClearReference(&entry->TargetDrvUp);
-            PhFree(entry);
+            PhDereferenceObject(entry->Name);   // HACK: resolver workers may be still running and need these values
```
jxy-s, line 1334: I wouldn't necessarily call this a hack. Maybe a better comment would be "do not NULL the value, the resolver thread takes a reference itself and might still be using it"?

plugins/ExtendedTools/objmgr.c
```diff
@@ -1158,64 +1300,65 @@ NTSTATUS EtEnumCurrentDirectoryObjects(
...
+            PhDereferenceObject(entry->TypeName);
+            PhClearReference(&entry->Target);
+            PhClearReference(&entry->TargetDrvLow);
+            PhClearReference(&entry->TargetDrvUp);
+            PhDereferenceObject(entry);
```
jxy-s, line 1339: Why not just let the final de-reference of the object entry clear/de-reference all the fields in the deletion routine?

plugins/ExtendedTools/objmgr.c
```diff
@@ -509,22 +537,23 @@ static BOOLEAN NTAPI EtEnumDirectoryObjectsCallback(
...
         }
     }
     else
     {
-        INT index;
-        POBJECT_ENTRY entry;
+        PET_OBJECT_ENTRY entry;
 
-        entry = PhAllocateZero(sizeof(ET_OBJECT_ENTRY));
+        entry = PhCreateAlloc(sizeof(ET_OBJECT_ENTRY));
```
jxy-s, line 554: 🤔 - I see we're using `ObDereferenceObject` on this elsewhere? But we're not constructing the object, just using a raw allocation?

jxy-s (changes requested):

plugins/ExtendedTools/objmgr.c
jxy-s: In order to do this you'll have to create an actual object type (`PhCreateObjectType`) instead of using `PhCreateAlloc`.

DartVanya: Changed to
```
EtObjectEntryType = PhCreateObjectType(L"ObjectItem", 0, EtpObjectEntryDeleteProcedure);
...
entry = PhCreateObjectZero(sizeof(ET_OBJECT_ENTRY), EtObjectEntryType);
```

plugins/ExtendedTools/objmgr.c
DartVanya: This code executes from UI thread only. For example, user press refresh button. Let's assume what resolver still working.

```
>	ExtendedTools.dll!EtObjectManagerFreeListViewItems(_ET_OBJECT_CONTEXT * Context) Строка 1317	C
 	ExtendedTools.dll!EtpObjectManagerRefresh(_ET_OBJECT_CONTEXT * Context) Строка 2367	C
 	ExtendedTools.dll!WinObjDlgProc(HWND__ * hwndDlg, unsigned int uMsg, unsigned __int64 wParam, __int64 lParam) Строка 2830	C
```
We send break command to current resolver thread, refresh tree and reselect previously active directory. After `EtpObjectManagerRefresh` calls `EtEnumCurrentDirectoryObjects`, which creates new resolver thread and store its `Break` switch pointer in BreakResolverThread.

```
Context->BreakResolverThread = (PBOOLEAN)&threadContext->Break;

PhCreateThread2(EtpTargetResolverThreadStart, threadContext);
```
Yes, it could possible be a race. Thanks for nit. I moved clearing of `context->BreakResolverThread` inside if block, so the pointer only be cleared if resolver wasn't imterrupted.
```
    if (status != STATUS_ABANDONED && !ReadAcquire8(&threadContext->Break))
    {
		...
        context->BreakResolverThread = NULL;
    }
```

plugins/ExtendedTools/objmgr.c
DartVanya: Implemented. Add BaseDirectory to ET_OBJECT_ENTRY.
`entry->BaseDirectory = PhReferenceObject(Context->CurrentPath);`
All needed resolver values now store inside ET_OBJECT_ENTRY and will dereference in object deletion procedure.

SystemInformer/findobj.c
DartVanya: Yes. hndllist.c and hndlprp.c will not show RefCnt in this case (RefCnt can be zero and it is valid value, but not ULONG_MAX)
```
            case PHHNTLC_POINTERCNT:
                if (handleItem->RefCnt != ULONG_MAX)
                    PhInitializeStringRefLongHint(&getCellText->Text, handleItem->PointerCountString);
```

jxy-s: A handle itself doesn't have a pointer count though? I guess I'm not clear on why this still here since we won't show anything?

DartVanya: If driver decode pointer count for us we can show handle pointer count. I removed driver changes, so now it will show nothing.

jxy-s: Understood. The "reference count" in the handle isn't really a reference count, it's very poorly named. It would be better described as a "use count". Even then it's an inverted value... Please just remove the dead code for now until we determine how to handle that.

plugins/ExtendedTools/objprp.c
jxy-s: Let me think about this a bit more. We can't guarantee we have an exclusive handle to something. And in many cases there will be multiple handles open to the same thing on a system. Then the cases where you could get the true reference wouldn't be many... Sorry to continue to press this, but let's please remove the code that doesn't have a clear accurate path forward for the time being. Showing inaccurate or partial information can be confusing or misleading (current case with the displayed pointer count).

DartVanya: 7f6029507d943847eecb3b84f7510f43eef0818e
I decide to hide reference count row from object properties window because this value does not currently contain useful and accurate information.
`PhRemoveListViewItem(context->ListViewHandle, PH_PLUGIN_HANDLE_GENERAL_INDEX_REFERENCES);`

jxy-s (approved):

PR #2317, opened 2024-12-03
fix native filtered handle enumeration

DartVanya:
PhEnumHandlesGeneric was changed in latest commits. Today I mention what SI crashed when I open handles of System (4). Driver was off.
I didn't understand that part of the code then, apparently it wasn't finished. I propose this solution for filtering handles from global dump.


dmex:
PhEnumHandlesGeneric was missing the line `if (handle->UniqueProcessId == ProcessId)` and has been fixed.

PR #2318, opened 2024-12-03
fix handle security if object name cannot be retrieved

DartVanya:
If caller passes a NULL name string to PhEditSecurity (e.g. hndlrpv cannot obtain object name), this will be null pointer dereference violation. Currently if you try to open security for any handle of PPL process (w/o driver), it leads to crash.


phlib/secedit.c
```diff
@@ -254,16 +254,16 @@ ISecurityInformation *PhSecurityInformation_Create(
     info->RefCount = 1;
 
     info->WindowHandle = WindowHandle;
-    info->ObjectName = PhCreateString(ObjectName);
-    info->ObjectType = PhCreateString(ObjectType);
+    info->ObjectName = ObjectName ? PhCreateString(ObjectName) : PhReferenceEmptyString();
+    info->ObjectType = ObjectType ? PhCreateString(ObjectType) : PhReferenceEmptyString();
```
dmex, line 258: Empty strings are not allowed per the API contract.

dmex (changes requested):

dmex:
PhEditSecurity explicitly defines those arguments with` _In_`  meaning those arguments are not optional and MSVC generates warnings when you're passing variables that are potentially null.

The caller must guarantee these parameters are not null... If they are null then don't call the function:

```
if (!(PhIsNullOrEmptyString(ObjectName) && PhIsNullOrEmptyString(ObjectType)))
{
   PhEditSecurity(...)
}
```

DartVanya:
Then we should use `PhGetStringOrEmpty(handleItem->ObjectName),`  there ?
https://github.com/winsiderss/systeminformer/pull/2317/files#diff-fd15e2cbc418345da148b647da9db28aaebd2f3f34bc120b12cd37046579ca30R643

SystemInformer/prpghndl.c
```diff
@@ -640,7 +640,7 @@ INT_PTR CALLBACK PhpProcessHandlesDlgProc(
 
                         PhEditSecurity(
                             PhCsForceNoParent ? NULL : hwndDlg,
-                            PhGetString(handleItem->ObjectName),
+                            PhGetStringOrEmpty(handleItem->ObjectName),
```
dmex, line 643: This was fixed upstream

plugins/ExtendedTools/objmgr.c
```diff
@@ -2264,7 +2264,7 @@ VOID NTAPI EtpObjectManagerObjectProperties(
     }
 
     objectContext.CurrentPath = PhReferenceObject(context->CurrentPath);
-    objectContext.Object = PhReferenceObject(Entry);
+    objectContext.Object = Entry;
```
dmex, line 2267: This can be rewritten with:

```cpp
 PhSetReference(&objectContext.CurrentPath, context->CurrentPath);
 PhSetReference(&objectContext.Object, Entry);
```

PhSetReference checks for null or references the object.

plugins/ExtendedTools/objmgr.c
```diff
@@ -2756,7 +2755,7 @@ VOID EtpObjectManagerCopyObjectAddress(
...
```
dmex, line 2758:
```cpp
PhSetReference(&objectContext.CurrentPath, context->CurrentPath);
PhSetReference(&objectContext.Object, Entry);
 ```

dmex (dismissed):

plugins/ExtendedTools/objmgr.c
DartVanya: In current implementation we cannot reference for `Entry` here, because it can be fake Directory object on stack. `ET_CREATE_DIRECTORY_OBJECT_ENTRY` macro fill necessary fields of `ET_OBJECT_ENTRY` to work with directory through common functions. Currently, the tree only holds directory string in item lParam, but not and object entry.
I added this code by mistake.

jxy-s (approved):

PR #2322, opened 2024-12-05
fix PhOpenDriver regression from 95bc2fe

DartVanya:
Directory and Driver objects cannot be opened with OBJ_EXCLUSIVE, system calls returns STATUS_INVALID_PARAMETER even currently is zero opened handles for this object. This breaks both PhOpenDriver and PhOpenDriverByBaseAddress.
https://github.com/winsiderss/systeminformer/commit/95bc2fe8c1a077aeb2103c076e3991683294d4fc#diff-7b1df2858b7dc31ac554743937d5e518d03d6957113a83969968d4b1626096c1R5621

### DIFF
--- a/phlib/native.c
+++ b/phlib/native.c
@@ -5618,7 +5618,7 @@ NTSTATUS PhOpenDriverByBaseAddress(
     InitializeObjectAttributes(
         &objectAttributes,
         &objectDirectoryName,
-        OBJ_EXCLUSIVE,
+        0,
         NULL,
         NULL
         );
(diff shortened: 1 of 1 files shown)


DartVanya:
> I'll partial revert the commit and merge with some other changes 👍

And I think what there is missing line, because now SI crashing if warning enabled and you try to start service (context->Object null pointer dereference).
https://github.com/winsiderss/systeminformer/blob/master/SystemInformer/actions.c#L3954
`context->Object = NumberOfServices == 1 ? L"the selected service" : L"the selected services";`

PR #2360, opened 2025-01-03
fix crash dialog dismiss behavior

ge0rdi:
When crash dialog was dismissed (escape, close dialog), `PhShowTaskDialog` returned `IDCANCEL` which value (2) was the same as value for "Normal" dump button.

It was probably not expected to create crash dump when user dismissed the dialog.

So we will use distinct values for custom dialog buttons.

### DIFF
--- a/SystemInformer/main.c
+++ b/SystemInformer/main.c
@@ -732,12 +732,12 @@ ULONG CALLBACK PhpUnhandledExceptionCallback(
         TASKDIALOGCONFIG config = { sizeof(TASKDIALOGCONFIG) };
         TASKDIALOG_BUTTON buttons[6] =
         {
-            { 1, L"Full\nA complete dump of the process, rarely needed most of the time." },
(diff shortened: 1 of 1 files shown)


ge0rdi:
One more thing related to crash dialog.
It seems many users create _mini dumps_ that contain just bare amount of information.

Shouldn't we somehow encourage them to create at least _normal_ dump?
We can make that button _default_.
Or add information that this is _preferred_.

What do you think?

jxy-s:
> One more thing related to crash dialog. It seems many users create _mini dumps_ that contain just bare amount of information.
> 
> Shouldn't we somehow encourage them to create at least _normal_ dump? We can make that button _default_. Or add information that this is _preferred_.
> 
> What do you think?

I think the default is appropriate (exit without collecting a dump). I would prefer at least a normal dump from crashes. I think the encouragement "this dump file is the most useful" is probably reasonable as it is. I'm not sure if adding "preferred" is a substantial nudge over what is already there.

jxy-s (approved):

PR #2366, opened 2025-01-05
fix debug console

ge0rdi:
Before:
image
Also _stdin_ didn't work, so it was not possible to execute commands.

After:
image
Stdin now works well too.


dmex:
This was a regression from binmode. Fixed with commit 93387c44395eb8728acfa0a4724c6f9685580ce2

PR #2403, opened 2025-01-24
Update util.c

DavidXanatos:
every ware else we use narrow string in KphTracePrint, so lets be consistent.


jxy-s (approved):

jxy-s:
Cleaned up here: https://github.com/winsiderss/systeminformer/commit/389700f4aa7d7bda8929f64c14dd604ba25e73c5

PR #2455, opened 2025-03-14
Fix begin_private comment placement

m417z:
I'm not sure if that's the correct fix according to the `// begin_*` comments convention. Currently, the code looks as following:
```cpp
#if (PHNT_VERSION >= PHNT_WINDOWS_VISTA)
// ...
// begin_private
// ...
#endif // PHNT_VERSION >= PHNT_WINDOWS_VISTA
// ...
// end_private
```
Which doesn't follow proper nesting unlike in other places. This confuses one of my scripts that parses the headers.

### DIFF
--- a/phnt/include/ntrtl.h
+++ b/phnt/include/ntrtl.h
@@ -8856,12 +8856,12 @@ RtlQueryValidationRunlevel(
 // Private namespaces
 //
 
+// begin_private
+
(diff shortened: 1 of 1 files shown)


jxy-s (approved):
Thanks @m417z, LGTM. I'll pull this into the phnt mirror after it's merged.

PR #2474, opened 2025-03-28
phnt: Use const string pointers for input arguments

m417z:


jxy-s (approved):

jxy-s:
I'll fix the driver build after this is merged 👍.

PR #2475, opened 2025-03-28
Add AFD handle information

diversenok:
This pull request adds definitions for interacting with `\Device\Afd` (Ancillary Function Driver) and implements support for viewing various information about socket handles.

image

### DIFF
--- a/SystemInformer/findobj.c
+++ b/SystemInformer/findobj.c
@@ -1602,29 +1602,17 @@ INT_PTR CALLBACK PhFindObjectsDlgProc(
                         if (handleObjectNode->ResultType == HandleSearchResult)
                         {
                             PPH_HANDLE_ITEM handleItem;
-                            BOOLEAN isAlpcPort = FALSE;
 
                             handleItem = PhCreateHandleItem(&handleObjectNode->HandleInfo);
 
                             if (!PhIsNullOrEmptyString(handleObjectNode->TypeNameString))
-                            {
-                                isAlpcPort = PhEqualString2(handleObjectNode->TypeNameString, L"ALPC Port", TRUE);
-                                handleItem->TypeName = handleObjectNode->TypeNameString;
-                                PhReferenceObject(handleObjectNode->TypeNameString);
-                            }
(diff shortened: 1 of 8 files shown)


SystemInformer/findobj.c
```diff
@@ -1602,29 +1602,17 @@ INT_PTR CALLBACK PhFindObjectsDlgProc(
...
 
                             if (!PhIsNullOrEmptyString(handleObjectNode->TypeNameString))
-                            {
-                                isAlpcPort = PhEqualString2(handleObjectNode->TypeNameString, L"ALPC Port", TRUE);
-                                handleItem->TypeName = handleObjectNode->TypeNameString;
-                                PhReferenceObject(handleObjectNode->TypeNameString);
-                            }
+                                handleItem->TypeName = PhReferenceObject(handleObjectNode->TypeNameString);
 
-                            // Best object name for the ALPC port contains information about the connection
```
jxy-s, line 1616: This reason this exists this way is because the `BestObjectName` contains a verbose string in the list so it can be searched and identified. But when expanding the properties we don't want to overflow the `Name` information. So we use the `ObjectName` for `ALPC Port` instead of `BestObjectName`. It's a quirk and an ugly hack in this path, I'd like to clean up this path. But in the meantime your change of this causes this to regress:

image

diversenok: I suppose the ALPC case doesn't really concern me (except for maybe the fact that the dialog currently shows different names depending on where it's invoked from, which is odd). I rewrote this piece of code because the else block did not forward the original object name, which I need to identify socket handles. If you want to keep the old ALPC case logic, feel free to revert it. As long as it doesn't break anything 😄

jxy-s: Well, I bring it up because I think AFD should probably get similar treatment:

image

jxy-s: Will fix after I merge 👍

jxy-s (approved):
Properties name needs attention, and flag formatting should use `PhGetAccessString`. Will address after merge.

PR #2476, opened 2025-03-28
Fix RTL_CONSTANT_STRING for ANSI strings

m417z:
In https://github.com/winsiderss/systeminformer/commit/f92eab70830a5772d1d8ff89315cb032258405a8, `RTL_CONSTANT_STRING` was updated from:
```cpp
#define RTL_CONSTANT_STRING(s) { sizeof(s) - sizeof((s)[0]), sizeof(s), s }
```
to:
```cpp
#define RTL_CONSTANT_STRING(s) { sizeof((s)) - sizeof((s)[0]), sizeof((s)), (PWCH)(s) }
```
to improve compatibility with C++.

This breaks the macro for ANSI strings, which should be supported [according to MSDN](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlinitstring):
> The **RTL_CONSTANT_STRING** macro creates a string or Unicode string structure to hold a counted string.

This PR adopts @namazso's suggestion from https://github.com/winsiderss/phnt/issues/48.


namazso:
Note that this would require a bunch of STL headers as-is, you might want to just reimplement all the template stuff without STL. The `_t` versions also require C++14, while the `::type` is only C++11.

m417z:
Got rid of STL. That's a bit verbose, let me know if you can think of a simpler definition.

jxy-s:
Should the definition here just mirror what is already in the WDK?:
```c
#ifdef __cplusplus
extern "C++"
{
template <size_t N> char _RTL_CONSTANT_STRING_type_check(const char  (&s)[N]);
template <size_t N> char _RTL_CONSTANT_STRING_type_check(const WCHAR (&s)[N]);
// __typeof would be desirable here instead of sizeof.
template <size_t N> class _RTL_CONSTANT_STRING_remove_const_template_class;
template <> class _RTL_CONSTANT_STRING_remove_const_template_class<sizeof(char)>  {public: typedef  char T; };
template <> class _RTL_CONSTANT_STRING_remove_const_template_class<sizeof(WCHAR)> {public: typedef WCHAR T; };
#define _RTL_CONSTANT_STRING_remove_const_macro(s) \
    (const_cast<_RTL_CONSTANT_STRING_remove_const_template_class<sizeof((s)[0])>::T*>(s))
}
#else
char _RTL_CONSTANT_STRING_type_check(const void *s);
#define _RTL_CONSTANT_STRING_remove_const_macro(s) (s)
#endif
#define RTL_CONSTANT_STRING(s) \
{ \
    sizeof( s ) - sizeof( (s)[0] ), \
    sizeof( s ) / sizeof(_RTL_CONSTANT_STRING_type_check(s)), \
    _RTL_CONSTANT_STRING_remove_const_macro(s) \
}
```

m417z:
Done.

jxy-s (approved):

PR #2492, opened 2025-04-04
Add more AFD socket information

diversenok:
This pull requests moves existing socket handle information to a dedicated page and adds support for displaying more properties from AFD info classes and transport-level options (on socket, IPv4/IPv6, TCP, UDP, and Hyper-V levels).

image


dmex:
You can also show SIO_TCP_INFO:
https://learn.microsoft.com/en-us/windows/win32/api/mstcpip/ns-mstcpip-tcp_info_v0
SIO_TIMESTAMPING:
https://learn.microsoft.com/en-us/windows/win32/api/mstcpip/ns-mstcpip-timestamping_config

I wrote a plugin long ago to query SIO_TCP_INFO from handles but required DuplicateHandle -> WSAStartup  -> WSAIoctl -> WSACleanup to succeed (ws2_32.dll would also throw an assert in debug mode) but should be easier using PhAfdQueryOption.

dmex (approved):

diversenok:
Oh yeah, great idea. Just checked, `SIO_TCP_INFO` works perfectly fine with the driver IOCTL. I'll see where I can put it.

PR #2493, opened 2025-04-04
Fix the bitness of the dynamic relocations for 32-bit

T-RN-R:
This PR swaps the struct used for calculating the base and end of the relocations. Originally, for calculations in 32-bit images, the 64-bit structure was being used. This swaps it to the correct 32-bit version


T-RN-R:

dmex (approved):

dmex:
Good catch

PR #2495, opened 2025-04-05
Improve FQBN version formatting

diversenok:
This pull request adjusts version formatting in FQBN security attributes and claims. It decodes values such as `2814751014981775` into more meaningful `10.0.19041.4239`.


jxy-s (approved):

diversenok:
Oops, I just realized I used `%hd` instead of `%hu`. Can somebody please fix that?

jxy-s:
> Oops, I just realized I used `%hd` instead of `%hu`. Can somebody please fix that?

Yes, we should also try to stick to using `PH_FORMAT` in the future (where possible).

PR #2496, opened 2025-04-05
phnt: Use relative include paths

m417z:
Currently, phnt must be added to the include path and used as following:
```c
#include <phnt_windows.h>
#include <phnt.h>
```

This change allows to just drop the headers to any folder and use:
```c
#include "my_libraries/phnt/phnt_windows.h"
#include "my_libraries/phnt/phnt.h"
```

Without having to change compilation flags or project configurations

### DIFF
--- a/phnt/include/ntexapi.h
+++ b/phnt/include/ntexapi.h
@@ -7,7 +7,7 @@
 #ifndef _NTEXAPI_H
 #define _NTEXAPI_H
 
-#include <ntkeapi.h>
+#include "ntkeapi.h"
 
 typedef struct _TEB* PTEB;
 typedef struct _COUNTED_REASON_CONTEXT* PCOUNTED_REASON_CONTEXT;
--- a/phnt/include/ntpebteb.h
+++ b/phnt/include/ntpebteb.h
@@ -7,7 +7,7 @@
 #ifndef _NTPEBTEB_H
 #define _NTPEBTEB_H
 
-#include <ntsxs.h>
+#include "ntsxs.h"
 
 typedef struct _RTL_USER_PROCESS_PARAMETERS *PRTL_USER_PROCESS_PARAMETERS;
 typedef struct _RTL_CRITICAL_SECTION *PRTL_CRITICAL_SECTION;
--- a/phnt/include/ntpsapi.h
+++ b/phnt/include/ntpsapi.h
@@ -7,7 +7,7 @@
 #ifndef _NTPSAPI_H
 #define _NTPSAPI_H
 
-#include <ntpebteb.h>
+#include "ntpebteb.h"
 
 //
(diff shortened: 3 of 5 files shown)


oberrich:
Stream of thought: It would probably make more sense to just let phnt be it's own subproject with its own issue tracker; standalone version that lags behind could become main version and always be on master. phnt could be released with a semver.
This obviously adds an extra step in sysinformer dev process but there's ways to mitigate this.
We could also add tests via CI then, to, at the very least, ensure the headers at least compile and are expected to function properly.
Currently downstream is plagued by upstream issues, because phnt_nightly/rust bindings keep getting temporarily erroneous code; this would require me to add CI to the phnt_nightly branch, at which point the phnt project should just be standalone and consumed by system informer, imho.

dmex:
We can't use relative paths because there's been issues with Visual Studio where some functions have definitions in both the Windows SDK and PHNT and the PHNT types end up excluded from intellisense and navigation, but used during compile and nothing displayed by Visual Studio (Internet Explorer 2.0) making sense.

If you include headers from a static library as relative, then include the static library in another project which also include the same headers, the definitions and ifdefs used when compiling the static library have precedence over the same typedefs used by the current project and you get nonsense errors and weird bugs. For example, when the library has a define allowing inline typedefs while the current project doesn't or has a different value the Visual Studio UI shows everything using defines of the current project (nothing inline) but uses the defines of the static library (everything inline) during compile, then doesn't add any correct information to symbols.

Like these winsock includes and defines used here in the static lib of main application will completely override anything included by separate projects like NetworkTools:
https://github.com/winsiderss/systeminformer/blob/02c680b9fdf86f488541e03c712e3f1bdd3331bd/phlib/include/phnet.h#L34-L44

If you define something in NetworkTools it's displayed as enabled and information added to debug symbols but doesn't get used during compile 🤷‍♂️ 

>  This change allows to just drop the headers to any folder and use:

Adding /include to the settings is easy enough. If you have Vcpkg installed then adding phnt to the config.json is enough to automatically include the headers, you essentially have a relative include with Vcpkg since it automatically includes the directories.

> It would probably make more sense to just let phnt be it's own subproject with its own issue tracker; standalone version that lags behind could become main version and always be on master. phnt could be released with a semver.

PHNT is a separate project but git submodules/remote dependencies are not permitted:
https://github.com/winsiderss/phnt

dmex:
@m417z 

The `build\build_header.cmd` script merges every PHNT header into `\phnt\include\nt.h` and you can use it standalone without any other files or includes.

m417z:
Nice, this looks like an official version of @mrexodia's https://github.com/mrexodia/phnt-single-header.

I think it'd be a good idea to have the merged header available at https://github.com/winsiderss/phnt, so that users won't have to run the script manually.

PR #2504, opened 2025-04-11
phnt: Fix `Rtl(En/De)code(System/Remote)Pointer` SAL annotations

RatinCN:
Consistent with the Windows SDK (utilapiset.h), for example:
```C
WINBASEAPI
_Ret_maybenull_
PVOID
WINAPI
EncodePointer(
    _In_opt_ PVOID Ptr
    );
```

And `Rtl(En/De)codeRemotePointer` return `HRESULT`, which wrap `NTSTATUS` by using `HRESULT_FROM_NT` macro.


phnt/include/ntrtl.h
```diff
@@ -9527,10 +9527,11 @@ RtlComputeCrc32(
  * @sa https://learn.microsoft.com/en-us/previous-versions/bb432254(v=vs.85)
  */
 NTSYSAPI
+_Ret_maybenull_
```
dmex, line 9530: It used to be standard to add SAL above the type and after NTSYSAPI statements but a few parsers have been known to ignore anything when the first statement is declspec like NTSYSAPI 

for example the CRT and SDK have moved all the SAL annotations to the first line before anything that's used by a linker in the function signature (i.e. _ACRTIMP _CRTALLOCATOR are linker/declspec):
image

phnt/include/ntrtl.h
```diff
@@ -9555,10 +9557,11 @@ RtlDecodePointer(
  * @sa https://learn.microsoft.com/en-us/previous-versions/bb432255(v=vs.85)
  */
 NTSYSAPI
+_Ret_maybenull_
```
dmex, line 9560: Need to be the first line above NTSYSAPI

dmex (changes requested):

phnt/include/ntrtl.h
dmex: `_Ret_maybenull_` also need to include the _check_return_ otherwise it doesn't raise warnings:

```
_Check_return_ 
_Ret_maybenull_
```

jxy-s: I believe `_Ret_maybenull_` alone is correct here since the input is allowed to be null. The caller should already know the condition of the input and shouldn't be required to check the condition of the output. And it is meaningless to do so anyway since the pointer will be scrambled.

phnt/include/ntrtl.h
```diff
@@ -9541,10 +9542,11 @@ RtlEncodePointer(
  * @sa https://learn.microsoft.com/en-us/previous-versions/bb432242(v=vs.85)
  */
 NTSYSAPI
+_Ret_maybenull_
```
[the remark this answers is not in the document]
RatinCN: Committed.

phnt/include/ntrtl.h
RatinCN: Committed.

phnt/include/ntrtl.h
RatinCN: Yes, both of input pointer and output pointer can be `NULL`, so SAL just mark them can be null and no need to check them.

jxy-s (approved):

PR #2517, opened 2025-04-20
phnt: Fix `RtlCompactHeap` return type

RatinCN:
My RE result, ReactOS, NTinterlnals.net, Wine, ... all of them show the result is an `ULONG`.

### DIFF
--- a/phnt/include/ntrtl.h
+++ b/phnt/include/ntrtl.h
@@ -5721,7 +5721,7 @@ RtlExtendHeap(
     );
 
 NTSYSAPI
-SIZE_T
+ULONG
 NTAPI
 RtlCompactHeap(
     _In_ PVOID HeapHandle,


jxy-s (approved):

dmex (approved):

diversenok:
I disagree with this change; the disassembly for `RtlCompactHeap` uses 64-bit moves to and from the variable that later becomes the result.

Also, never use the NTinternals website for these things. Especially the memory management category. All definitions there are broken because they assume 32-bit pointers.  ReactOS isn't exactly trustworthy in this regard either.

jxy-s:
> I disagree with this change; the disassembly for `RtlCompactHeap` uses 64-bit moves to and from the variable that later becomes the result.
> 
> Also, never use the NTinternals website for these things. Especially the memory management category. All definitions there are broken because they assume 32-bit pointers.  ReactOS isn't exactly trustworthy in this regard either.
> 

I should have checked myself before accepting... I'll look when I have an opportunity.

jxy-s:
Confirmed correct return size is variable width based on the size of a register. `SIZE_T` is correct. I've reverted the commit: https://github.com/winsiderss/systeminformer/commit/f45008757c3d12dfd1ec0b571e3bc4316c98030f

RatinCN:
Sorry, I was carelessly mistaken, I checked that `SIZE_T` was right.

PR #2522, opened 2025-04-25
Improve LdrSystemDllInitBlock version checks

diversenok:
This pull request records how `PS_SYSTEM_DLL_INIT_BLOCK` changed over the years and marks which fields are safe to access at their current offsets for each version of Windows. It also updates `PhGetProcessSystemDllInitBlock` to perform necessary validation. The new parameter allows the callers to specify the furthest field they intend to access.


phlib/include/phnative.h
```diff
@@ -3821,7 +3821,8 @@ NTSTATUS
 NTAPI
 PhGetProcessSystemDllInitBlock(
     _In_ HANDLE ProcessHandle,
-    _Out_ PPS_SYSTEM_DLL_INIT_BLOCK SystemDllInitBlock
+    _In_ ULONG RequiredSize,
+    _Out_writes_bytes_(RequiredSize) PPS_SYSTEM_DLL_INIT_BLOCK SystemDllInitBlock
```
jxy-s, line 3825: Should this include `_Post_readable_size_(RequiredSize)`?

diversenok: Yes, good idea. Although, it makes me wonder why one doesn't automatically follow form the other.

jxy-s (approved):
I'm good with these changes. Seems like a decent pattern. But I'd like to get @dmex's input before merging.

phnt/include/ntldr.h
```diff
@@ -649,7 +649,7 @@ typedef struct _PS_MITIGATION_AUDIT_OPTIONS_MAP
 typedef struct _PS_SYSTEM_DLL_INIT_BLOCK
 {
     ULONG Size;
-    ULONG_PTR SystemDllWowRelocation;
+    ULONG_PTR SystemDllWowRelocation; // current offset since REDSTONE2
```
dmex, line 652:
```
// since RS2
```

phnt/include/ntldr.h
```diff
@@ -662,12 +662,12 @@ typedef struct _PS_SYSTEM_DLL_INIT_BLOCK
             ULONG Reserved : 31;
         };
     };
-    PS_MITIGATION_OPTIONS_MAP MitigationOptionsMap;
-    ULONG_PTR CfgBitMap;
+    PS_MITIGATION_OPTIONS_MAP MitigationOptionsMap; // current size since 20H1
```
dmex, line 665:
```
// since 20H1
```

phnt/include/ntldr.h
```diff
@@ -662,12 +662,12 @@ typedef struct _PS_SYSTEM_DLL_INIT_BLOCK
...
+    ULONG_PTR CfgBitMap; // current offset since 20H1
```
dmex, line 666: Which field is since 20H1?

phnt/include/ntldr.h
```diff
@@ -662,12 +662,12 @@ typedef struct _PS_SYSTEM_DLL_INIT_BLOCK
...
     ULONG_PTR CfgBitMapSize;
     ULONG_PTR Wow64CfgBitMap;
     ULONG_PTR Wow64CfgBitMapSize;
-    PS_MITIGATION_AUDIT_OPTIONS_MAP MitigationAuditOptionsMap; // REDSTONE3
+    PS_MITIGATION_AUDIT_OPTIONS_MAP MitigationAuditOptionsMap;
```
dmex, line 670: Why remove the version?

phnt/include/ntldr.h
```diff
@@ -677,19 +677,11 @@ typedef struct _PS_SYSTEM_DLL_INIT_BLOCK
     ULONG_PTR ScpArm64EcCfgCheckESFunction;
 } PS_SYSTEM_DLL_INIT_BLOCK, *PPS_SYSTEM_DLL_INIT_BLOCK;
 
-// rev
-#if (PHNT_VERSION >= PHNT_WINDOWS_10)
+#if (PHNT_VERSION >= PHNT_WINDOWS_8)
```
dmex, line 680: Microsoft define the type in the Windows SDK 10.0.10240.0/um/minwin/wow64t.h

```
#if defined(_NTDLLBUILD_)
extern PS_SYSTEM_DLL_INIT_BLOCK LdrSystemDllInitBlock;
#else
extern NTSYSAPI PS_SYSTEM_DLL_INIT_BLOCK LdrSystemDllInitBlock;
#endif // defined(_NTDLLBUILD_)
```

dmex (changes requested):

phnt/include/ntldr.h
diversenok: Interesting, didn't know it was there. Either way, the export and corresponding type exists in ntdll and its symbols all the way since Windows 8 RTM.

phnt/include/ntldr.h
diversenok: Do we not want to differentiate between fields that are new to a specific OS version vs. fields that existed before but changed their offset?

phnt/include/ntldr.h
diversenok: That depends on what we want to mark as "since". `PS_MITIGATION_OPTIONS_MAP` contained a two element array before 20H1, now it's three. So `MitigationOptionsMap` used to start at the same offset but was smaller. Thus, `current size since 20H1`. And `CfgBitMap` used to start earlier, thus, `current offset since 20H1`.

phnt/include/ntldr.h
diversenok: Because (at its current offset) the field is usable only since the same version (20H1) as a marked a few fields above. The full comment would be `since REDSTONE3, current offset since 20H1`

phnt/include/ntldr.h
dmex: I think two separate structures are needed with a V1/2 suffix since this and the below MitigationAuditOptionsMap for RS3 are incompatible with later versions?

diversenok: Yes, sure. If we want to support as many fields as possible on as many OS versions as possible, three revisions seem like the minimum (since MS broke offsets twice). One structure for 8-RS1, another for RS2-19H2, and the newest for 20H1 and above. I can add them to `ntldr.h` and then hide as an implementation detail in `PhGetProcessSystemDllInitBlock` by making sure it always converts them to the latest format before returning. Does that sound good?

dmex: Yes

diversenok:
I added different revisions for `PS_SYSTEM_DLL_INIT_BLOCK` and a function that converts them to the latest layout. `PhGetSystemDllInitBlock` and `PhGetProcessSystemDllInitBlock` now return a version-indpendent copy for the current and other process's blocks, respectively. The callers can use `RTL_CONTAINS_FIELD` on the result instead of version checks since the code updates the size field according to the other field's availability.

I also removed `LdrControlFlowGuardEnforcedWithExportSuppression` due to the many issues it had:
1. It didn't read correct fields on older versions (assumed the latest layout)
2. It didn't read correct fields on 32-bit versions (wrong use of `ULONG_PTR` instead of `ULONG64` in the underlying types)
3. It didn't test the correct mitigation policy bit (`3` instead of `0x30000000000`)
4. It used to import `LdrSystemDllInitBlock` in a non-delayed fashion
5. Attempting to use it via `LdrControlFlowGuardEnforcedWithExportSuppression_Import()` never worked because there is no ntdll export with the same name
6. Even after fixing all these issues and re-implementing it using `LdrSystemDllInitBlock_Import()` with proper version handling, turns out that its only consumer - `PhGetDllBaseProcedureAddress` - will deadlock while attempting to call `LdrSystemDllInitBlock_Import()` due to lock reentrance problems.

TLDR; the CFG export suppression check in `PhGetDllBaseProcedureAddress` can now actually return TRUE, so watch out for potential side effects.

phnt/include/ntldr.h
```diff
@@ -1384,57 +1448,20 @@ LdrSetImplicitPathOptions(
...
-    return LdrSystemDllInitBlock.CfgBitMap && (LdrSystemDllInitBlock.Flags & 1) == 0;
-}
-#else
-// rev
+#if (PHNT_VERSION >= PHNT_WINDOWS_10RS3)
+// private
 /**
  * The LdrControlFlowGuardEnforced function checks if Control Flow Guard is enforced.
  *
  * @return BOOLEAN TRUE if Control Flow Guard is enforced, FALSE otherwise.
```
jxy-s, line 1456: The documentation needs updated. 

Looks like `al` isn't in routine or by the call sites in ntdll, they all operate with `eax`. So I agree that this probably shouldn't have been `BOOLEAN`. OTOH `BOOLEAN` seems convenient since only the LSB is ever relevant according to the disassembly :shrug:.

phnt/include/ntldr.h
```diff
@@ -1384,57 +1448,20 @@ LdrSetImplicitPathOptions(
...
-FORCEINLINE
-BOOLEAN
-NTAPI
-LdrControlFlowGuardEnforcedWithExportSuppression(
-    VOID
-    )
-{
-    return LdrSystemDllInitBlock.CfgBitMap
-        && (LdrSystemDllInitBlock.Flags & 1) == 0
-        && (LdrSystemDllInitBlock.MitigationOptionsMap.Map[0] & 3) == 3; // PROCESS_CREATION_MITIGATION_POLICY_CONTROL_FLOW_GUARD_EXPORT_SUPPRESSION
```
jxy-s, line 1435: Yep, this should have been `BYTE5(LdrSystemDllInitBlock.MitigationOptionsMap.Map[0])`.
```
.text:0000000180021A44                 mov     rcx, cs:LdrSystemDllInitBlock.MitigationOptionsMap.Map
.text:0000000180021A4B                 shr     rcx, 28h
.text:0000000180021A4F                 and     cl, 3
.text:0000000180021A52                 cmp     cl, 3
```

phlib/mapldr.c
```diff
@@ -1052,6 +1052,193 @@ PVOID PhGetLoaderEntryDllBase(
...
+ * \param[in] Source A buffer returned or copied from LdrSystemDllInitBlock.
+ * \param[out] Destination A buffer for storing a version-independent copy of the structure.
+ */
+VOID PhCaptureSystemDllInitBlock(
+    _In_ PVOID Source,
+    _Out_ PPS_SYSTEM_DLL_INIT_BLOCK Destination
+    )
+{
+    ULONG sourceSize;
+    PS_SYSTEM_DLL_INIT_BLOCK copy = { 0 };
```
jxy-s, line 1067: I think a copy could be avoided here by zeroing the output `Desintation` and just populating that instead of the intermediate `copy`.

jxy-s (approved):
I'll want to run through some tests on my machine with ES forced. I've had a number of problems in the past with the chicken and egg in our special loader logic.

I don't mine doing the minor cleanup myself, changes look fine to me.

PR #2529, opened 2025-04-30
Rework the COM apartment column

diversenok:
This pull request improves how the thread page displays information about COM state. First, the PR renames the `COM apartment` column to `COM flags`, preserving its functionality. Second, it introduces a new `COM apartment` column that displays the apartment type and qualifier similar to [CoGetApartmentType](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cogetapartmenttype) plus the number of reentrant COM initializations on the thread (if more than one).

image

The new column uses three sources of information:
1. The same OLE TLS flags as the old column.
2. The COM initialization counter from the same OLE TLS structure.
3. The process-wide MTA usage information, which helps identify existence of MTA in the process and correctly classify threads without an explicit apartment into implicit MTA. It uses `CoGetMTAUsageInfo` to retrieve pointers to global variables in `combase.dll` which we can read cross-process.
4. Searching for the main STA window to differentiate between main and non-main STAs. This step uses `NtUserFindWindowEx` to iterate over message-only windows with `OleMainThreadWndClass` and `OleMainThreadWndName` and checks their owner via `NtUserQueryWindow`. Ideally, we would read the global combase variable for that (the same way `CoGetApartmentType` does it), but I cannot find a simple way of locating it. Performance-wise, it might not be the best solution, but even the worst case scenario takes less than a millisecond on average my machine. I also looked into using `NtUserBuildHwndList` to filter by TID before testing the window class and name, but it doesn't seems to include message-only windows. Either way, this code path triggers only for STA threads and only when the column is enabled.


phlib/nativethread.c
```diff
@@ -811,32 +811,204 @@ NTSTATUS PhGetThreadLastStatusValue(
...
+#endif
+
+    if (PhBeginInitOnce(&initOnce))
+    {
+        PVOID combase;
+        PMTA_USAGE_GLOBALS (WINAPI* CoGetMTAUsageInfo_I)(VOID);
+
+        if (combase = PhGetLoaderEntryDllBaseZ(L"combase.dll"))
+        {
+            // combase exports CoGetMTAUsageInfo as ordinal 70 since Windows 8
```
jxy-s, line 854: Should this be guarded by `>= WINDOWS_8`?

phlib/nativethread.c
```diff
@@ -811,32 +811,204 @@ NTSTATUS PhGetThreadLastStatusValue(
...
+}
+
+/**
+ * Determines if a thread belongs to the main STA by checking if it owns a main STA window.
+ *
+ * \param[in] ThreadHandle A handle to the thread. The handle must have THREAD_QUERY_LIMITED_INFORMATION access.
+ *
+ * \return Whether the thread owns a main STA window.
+ */
+BOOLEAN PhIsThreadMainSTA(
```
dmex, line 969: It might be better to have a single function and enum to check against?

https://learn.microsoft.com/en-us/windows/win32/cossdk/threading-model-attribute

```
    Main Thread Apartment
    Single Thread Apartment
    Free Thread Apartment
    Neutral Apartment
    Any Apartment
````


PhIsThreadApartment(_In_ HANDLE ThreadHandle, enum value to compare?) similar to `PhGetThreadApartment`

**Review by dmex — CHANGES_REQUESTED** [review]

File: phlib/nativethread.c (hunk 1 above)
Comment (diversenok) [reply phlib/nativethread.c:854]: `combase.dll` itself exists since Windows 8, so `CoGetMTAUsageInfo` is always available there. But yeah, I'll add a version check to avoid unsuccessful loading attempts on earlier versions.

File: phlib/nativethread.c [hunk 3]
```diff
@@ -811,32 +811,204 @@ NTSTATUS PhGetThreadLastStatusValue(
...
+    }
+
+    if (!NtUserFindWindowEx_I || !NtUserQueryWindow_I)
+        return FALSE;
+
+    hwnd = NULL;
+    do
+    {
+        // Find the next main STA window
+        hwnd = NtUserFindWindowEx_I(
```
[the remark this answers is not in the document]
diversenok: I don't think it's that recent anymore (RS1 is from 2016) 😄 
I'll update it.

phlib/nativethread.c
diversenok: It's a helper function for telling apart main and non-main STA since we cannot do that via OLE TLS flags. I probably should've used a `Php` prefix and avoid putting it in the header. I think I'll just merge it into `PhGetThreadApartment` altogether.

phlib/nativethread.c
```diff
@@ -811,32 +811,204 @@ NTSTATUS PhGetThreadLastStatusValue(
...
+            );
+
+        // Check if it belongs to the specified thread ID
+    } while (hwnd && NtUserQueryWindow_I(hwnd, WindowThread) != (ULONG_PTR)ThreadId);
+
+    return !!hwnd;
+}
+
+/**
+ * Determines if a thread belongs to the main STA by checking if it owns a main STA window.
```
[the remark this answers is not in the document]
diversenok: I'll update the comments to be more explicit

phlib/nativethread.c
```diff
@@ -868,36 +965,172 @@ NTSTATUS PhGetThreadApartmentState(
...
+        sizeof(OLETLSFLAGS),
+        NULL
+        );
+
+    if (!NT_SUCCESS(status))
+        return status;
+
+    if (ComInits)
+    {
+        PVOID comInitsOffset;
```
dmex, line 1000: phlib\nativethread.c(1011,1): error C4700: uninitialized local variable 'comInitsOffset' used

dmex (changes requested):
Build failed due to phlib\nativethread.c(1011,1): error C4700: uninitialized local variable 'comInitsOffset' used

phlib/nativethread.c
diversenok: Oops, sorry, that was an #ifdef'ed typo affecting 32-bit builds. Fixed.

dmex (approved):

phlib/nativethread.c
jxy-s: Thanks, I was mostly concerned if the ordinal was in use prior to Windows 8. I've seen ordinal change and get reused, it's rare, but happens.

PR #2547, opened 2025-05-07
Update ntldr.h

diversenok:
This pull request adds some missing types, updates function prototypes to use official parameter names, and adjusts minimal import versions in ntldr. It also removes several functions that don't have matching ntdll exports, namely `LdrRelocateImage`, `LdrRelocateImageWithBias`, `LdrVerifyMappedImageMatchesChecksum`, and `LdrDllRedirectionCallback`.

Additionally, I'm not sure what to do with `RtlpScpCfgNtdllExports`. It looks like we already have the official types for it (`RTL_SCP_CFG_NTDLL_EXPORTS` and co.) in `ntmmapi.h` which can replace the reversed version of the same type `RTL_SCPCFG_NTDLL_EXPORTS` from `ntldr.h`. I don't know if you want to move the function to `ntmmapi.h` (closer to the types, so it can use them) or move both into `ntrtl.h`, or something else.


jxy-s (changes requested):
I'm seeing a lot of change here that needs more discussion and review. There are clearly some existing errant things and the changes here make some improvement. But I have more questions that outright agreement. I have not had time to get through all the changes. I will need to find time to align/convince myself of these changes.

phnt/include/ntldr.h
```diff
@@ -52,56 +55,128 @@ typedef enum _LDR_DDAG_STATE
...
+        struct
+        {
+            ULONG ForwarderLink : 1;
+            ULONG SpareFlags : 2;
+        };
+    };
+} LDRP_DEPENDENCY, *PLDRP_DEPENDENCY;
 
+// private
 typedef enum _LDR_DLL_LOAD_REASON
```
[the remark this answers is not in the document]
diversenok: A lot of places in phnt imply that `since` comments also apply to all fields/values defined below (until the end of the type), so I wanted to move `LoadReasonUnknown` to the front, outside of the `// since WIN11` effect.

As for explicitly specifying enum values, it is indeed superfluous for the compiler and the IDE. But I find it more convenient when reading/comparing header files. I can restore it and refrain from using explicit enum values in the future.

phnt/include/ntldr.h
```diff
@@ -1378,12 +1594,12 @@ NTSYSAPI
 NTSTATUS
 NTAPI
 LdrResolveDelayLoadsFromDll(
-    _In_ PVOID ParentModuleBase,
-    _In_ PCSTR TargetDllName,
+    _In_ PVOID ImportModuleBase,
+    _In_ PCSTR ExportDllName,
```
dmex, line 1598: ImportModuleBase needs to include DllBase or BaseAddress instead of ModuleBase

ExportDllName should probably remain TargetDllName? The first variable named Import, and the second variable named as Export is implying you pass the address of an image with imports as the first and the address of an image with exports as second?

phnt/include/ntldr.h
```diff
@@ -1075,86 +1266,101 @@ LdrNameOrIdFromResourceEntry(
...
     );
 
+// private
 NTSYSAPI
 NTSTATUS
 NTAPI
 LdrFindEntryForAddress(
-    _In_ PVOID DllHandle,
-    _Out_ PLDR_DATA_TABLE_ENTRY *Entry
+    _In_ PVOID Address,
```
dmex, line 1286: Address should be BaseAddress since it's the same as NtQueryVirtualMemeory: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-zwqueryvirtualmemory

phnt/include/ntldr.h
```diff
@@ -1331,40 +1549,38 @@ NTSYSAPI
...
+ * @param Flags Flags for the operation.
  * @return The address of the import, or the failure stub for it.
  * @remarks https://learn.microsoft.com/en-us/windows/win32/devnotes/resolvedelayloadedapi
  */
 NTSYSAPI
 PVOID
 NTAPI
 LdrResolveDelayLoadedAPI(
-    _In_ PVOID ParentModuleBase,
+    _In_ PVOID ImportModuleBase,
```
dmex, line 1575: ImportModuleBase needs to include DllBase or BaseAddress instead of ModuleBase

dmex (changes requested):

diversenok:
I did not invent any of these names. Everything marked with `// private` should match the official ones 🤷

phnt/include/ntldr.h
```diff
@@ -283,56 +367,62 @@ LdrGetDllHandleByName(
...
+// private
 NTSYSAPI
 NTSTATUS
 NTAPI
 LdrGetDllPath(
-    _In_  PCWSTR DllName,
-    _In_  ULONG  Flags, // LOAD_LIBRARY_SEARCH_*
-    _Out_ PWSTR* DllPath,
-    _Out_ PWSTR* SearchPaths
+    _In_  PCWSTR RootDllName,
```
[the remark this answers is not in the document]
diversenok: Do we want to to that with all `PCWSTR` parameters? Also, `PCWSTR` already includes a `_Null_terminated_` annotation

phnt/include/ntldr.h
```diff
@@ -820,29 +903,31 @@ NTSTATUS
 NTAPI
 LdrGetFileNameFromLoadAsDataTable(
     _In_ PVOID Module,
-    _Out_ PVOID *pFileNamePrt
+    _Out_ PVOID *FilePath
```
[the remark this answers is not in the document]
diversenok: Strangely, it's officially defined as such. Given the implementation, it should've been `PWSTR*`, so I'll update it.

phnt/include/ntldr.h
```diff
@@ -8,31 +8,34 @@
...
 
-typedef _Function_class_(LDR_INIT_ROUTINE)
-BOOLEAN NTAPI LDR_INIT_ROUTINE(
+// private
+typedef _Function_class_(DLL_INIT_ROUTINE)
+BOOLEAN NTAPI DLL_INIT_ROUTINE(
     _In_ PVOID DllHandle,
     _In_ ULONG Reason,
-    _In_opt_ PVOID Context
+    _In_opt_ PCONTEXT Context
```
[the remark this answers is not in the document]
diversenok: You know what, I pondered it a bit more and I think it should be `PCONTEXT`. The symbols clearly define the argument as `PCONTEXT`, so any DLL can reasonably expect it to be a context. The fact that it's possible to manually call the entry point and pass it custom data doesn't change the API contract.

phnt/include/ntldr.h
```diff
@@ -852,28 +937,51 @@ NTAPI
...
+{
+    ULONG_PTR ResourceIdPath[LdrResourceIdCount];
+    struct
+    {
+        PCWSTR Type; // RT_*
+        PCWSTR Name; // string or MAKEINTRESOURCE
+        PCWSTR Language; // string or LANGID
+        ULONG_PTR Item; // e.g., MessageId
+    };
+} LDR_RESOURCE_ID, *PLDR_RESOURCE_ID;
```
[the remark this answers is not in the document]
diversenok: > If the function is just using ULONG_PTR[] and the array size then we have to use it the same way.

It looks like it. It's a bit tricky, because I can only see that `ResourceIdPath` is not a struct pointer and that it is equivalent to `CONST UINT64*` on 64-bits and `CONST UINT*` on 32-bits. It might also be any equivalent typedef as these would be indistinguishable. I just dug a bit further and found a promising candidate: a typedef called `LDR_EXTERNAL_DLL_PATH` that maps precisely to `UINT_PTR`. There is no corresponding `PLDR_EXTERNAL_DLL_PATH` typedef, so I assume Microsoft uses
`_In_reads_(ResourceIdPathLength) CONST LDR_EXTERNAL_DLL_PATH ResourceIdPath[]` for the function parameter, plus some macros that allow more meaningful access to specific array elements. That would match perfectly with my observations. So I guess I should indeed remove these helper types and add a few macros for indexing instead.

phnt/include/ntldr.h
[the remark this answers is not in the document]
jxy-s: Huh, so it does. IMO having it on the SAL without having to "peek through" or know is most explicit/convenient. I guess I don't really care either way as along as prefast is happy.

jxy-s (approved):
> I did not invent any of these names. Everything marked with // private should match the official ones 🤷

I'm choosing to approve the PR since the changes seem functionally fine to me and I don't think the PR should be blocked just for naming disagreements.

That said, whether in this PR or as a follow up - I do think we should go back and clean up the parameter naming using reasonable judgement for consistently and clarity. IMO some of the naming is regressing a bit in this PR. I do not believe that Microsoft got the parameter naming right to begin with, parameter names (and to some degree types) are less set in stone compared to structure names/parameters - phnt is a good place (using our best judgement) to clean this up.

diversenok:
I think the official parameter names define the ground truth and we shouldn't deviate much from them. Yes, it's clear that Microsoft haven't established a consistent naming convention, and yes, we have a good position to fix big blunders. But doing renames like `ModuleBase` --> `DllBase` seems like micro-managing the task. After all, we don't go and redefine things from the SDK just because we disagree with their naming scheme. And I find official names in the definitions more valuable than the custom ones because they are reproducible and rediscoverable.

If I understand it correctly, the main concern is that some names might be confusing or misleading. The good place to resolve this confusion is in the docstring comments that @dmex introduces throughout phnt. If anything, that's the place where such comments can provide the most value.

phlib/mapldr.c
```diff
@@ -682,18 +682,18 @@ NTSTATUS PhLoadResource(
...
-    resourceInfo.Name = (ULONG_PTR)Name;
-    resourceInfo.Language = MAKELANGID(LANG_NEUTRAL, SUBLANG_NEUTRAL);
+    resourceIdPath[LDR_RESOURCE_PATH_TYPE] = (ULONG_PTR)Type;
+    resourceIdPath[LDR_RESOURCE_PATH_NAME] = (ULONG_PTR)Name;
+    resourceIdPath[LDR_RESOURCE_PATH_LANGUAGE] = MAKELANGID(LANG_NEUTRAL, SUBLANG_NEUTRAL);
 
     __try
     {
-        status = LdrFindResource_U(DllBase, &resourceInfo, RESOURCE_DATA_LEVEL, &resourceData);
+        status = LdrFindResource_U(DllBase, resourceIdPath, RTL_NUMBER_OF(resourceIdPath), &resourceData);
```
jxy-s, line 696: I don't think these can be in an arbitrary order? There is an explicit leveling to locating a resource. So having the `LDR_RESOURCE_INFO` be explicit seems most convenient to me. I'd also like information on where/why the name `LDR_EXTERNAL_DLL_PATH` was found/chosen. The "external" seems like an odd choice, the resource could be internal to my own (and even current) module.

phnt/include/ntldr.h
```diff
@@ -852,28 +937,26 @@ NTAPI
 LdrAccessResource(
     _In_ PVOID DllHandle,
     _In_ PIMAGE_RESOURCE_DATA_ENTRY ResourceDataEntry,
-    _Out_opt_ PVOID *ResourceBuffer,
-    _Out_opt_ ULONG *ResourceLength
+    _Out_opt_ PVOID* Address,
+    _Out_opt_ PULONG Size
```
jxy-s, line 941: I get that this leaked in NT5, but I find this to be a regression in clarity. :shrug:

phnt/include/ntldr.h
```diff
@@ -883,163 +966,246 @@ NTSTATUS
 NTAPI
 LdrFindResource_U(
     _In_ PVOID DllHandle,
-    _In_ PLDR_RESOURCE_INFO ResourceInfo,
-    _In_ ULONG Level,
-    _Out_ PIMAGE_RESOURCE_DATA_ENTRY *ResourceDataEntry
+    _In_reads_(ResourceIdPathLength) LDR_EXTERNAL_DLL_PATH ResourceIdPath[],
```
jxy-s, line 969: If we really want to eliminate `LDR_RESOURCE_INFO`. I'd change this to just be `PULONG_PTR` and leave out `LDR_EXTERNAL_DLL_PATH`.

phnt/include/ntldr.h
```diff
@@ -356,21 +447,23 @@ LdrGetProcedureAddressEx(
...
 
+// private
 NTSYSAPI
 NTSTATUS
 NTAPI
 LdrGetKnownDllSectionHandle(
     _In_ PCWSTR DllName,
-    _In_ BOOLEAN KnownDlls32,
-    _Out_ PHANDLE Section
+    _In_ BOOLEAN SearchKnownDlls32,
```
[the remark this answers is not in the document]
diversenok: This parameter used to control whether the function opens a handle from `KnownDlls` (FALSE) or `KnownDlls32` (TRUE) directory. Either way, it's unused on modern versions.

phnt/include/ntldr.h
diversenok: The address doesn't have to be the base address though; it could be anything within the DLL limits

phlib/mapldr.c
diversenok: I'll restore it back to when there was a helper union. As for `LDR_EXTERNAL_DLL_PATH`, see https://github.com/winsiderss/systeminformer/pull/2547#discussion_r2078760564

phnt/include/ntldr.h
diversenok: Indeed, each index has a separate meaning. I initially used `PULONG_PTR` but kept a helper type (with some adjustments to make it a union), but I got the impression that dmex didn't like the new type. I'll remove `LDR_EXTERNAL_DLL_PATH` in favor or `PULONG_PTR` and add back the helper type

phnt/include/ntldr.h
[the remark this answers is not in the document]
dmex: > The symbols clearly define the argument as PCONTEXT

The ntdll symbols might define the argument as PCONTEXT but it's not the only user and others don't pass CONTEXT.

> The fact that it's possible to manually call the entry point and pass it custom data doesn't change the API contract.

The only reason the LoadLibraryEx(DONT_RESOLVE_DLL_REFERENCES) and LdrLoadDll(IMAGE_FILE_EXECUTABLE_IMAGE) flags exist is for passing data via the Context parameter.

Issue #2508, opened 2025-04-13
System informer crashes when downloading updates.


kotenok2000:
### Brief description of your issue

Sometimes sy6stem informer crashes when it downloads an update.
When download reaches 100% it crashes.

### Environment (optional)

```shell
3.2.25099.1530
Microsoft Windows [Version 10.0.26100.3775]
```

kotenok2000:
[SystemInformer_DumpFile_TTILSVIQODYKNOM.dmp](https://github.com/user-attachments/files/19722578/SystemInformer_DumpFile_TTILSVIQODYKNOM.dmp)

dmex:
There's some issue with bcrypt.dll on your machine. It's internal to the library and not something we can fix.

MagicAndre1981:
Today I got my first SI crash for a while while updater is involded:

```
0x0000008fce0fe918 : 0x00007ff9c4a24b5d : ntdll!RtlGetExtendedContextLength2+0x3d
0x0000008fce0fe928 : 0x0000008fce0febc0 :  !du "\Device\"
0x0000008fce0fe968 : 0x00007ff9c49ee6ef : ntdll!RtlDispatchException+0xef
0x0000008fce0fe978 : 0x0000008fce0febc0 :  !du "\Device\"
0x0000008fce0fe990 : 0x0065002e00320077 :  !du "w2.exe"
0x0000008fce0fe9c0 : 0x0000008fce0fef88 :  !du "indows\system32\svchost.exe"
0x0000008fce0fe9c8 : 0x0000008fce0fef90 :  !du "ws\system32\svchost.exe"
0x0000008fce0fe9d8 : 0x00007ff9c4b5ebf8 : "ntdll!__PchSym_ <PERF> (ntdll+0x18ebf8)"
0x0000008fce0fe9e8 : 0x00007ff9c4a5f770 : ntdll!_C_specific_handler
0x0000008fce0fe9f8 : 0x00007ff9c4a2af38 : ntdll!RtlUserThreadStart+0x28
0x0000008fce0fea08 : 0x00007ff9c49d0000 : "ntdll!LdrpGetModuleName <PERF> (ntdll+0x0)"
0x0000008fce0fea20 : 0x00007ff9c4a2af38 : ntdll!RtlUserThreadStart+0x28
0x0000008fce0fea28 : 0x00007ff9c49d0000 : "ntdll!LdrpGetModuleName <PERF> (ntdll+0x0)"
[...]

PR #2548, opened 2025-05-07
Fix PhEnumDirectoryFileEx

ge0rdi:
SearchPattern parameter is optional, so we should use it only if valid one was provided.

This should fix crashing in update check (among others?) mentioned in https://github.com/winsiderss/systeminformer/issues/2508#issuecomment-2857683017.


dmex:
Fixed with https://github.com/winsiderss/systeminformer/commit/594bd8652237b5a3aa89759c3f37c4c69331f910

PR #2551, opened 2025-05-09
Rework API Set resolution

### LINKED ISSUE #2519: The layout of version 2 of the structure API_SET_NAMESPACE is missing
### Brief description of your issue

in the definition of _PEB, 

    //
    // Pointer to the API Set Schema.
    //
    PAPI_SET_NAMESPACE ApiSetMap;

Windows 7 uses version 2 (instead of 6) of API_SET_NAMESPACE.

the version 2 layout is described at (and named PAPISETMAP instead of API_SET_NAMESPACE): 
https://blog.quarkslab.com/runtime-dll-name-resolution-apisetschema-part-ii.html

I've personally verified that the layout given at the above link is correct for Windows 7 SP1.  I do not know if it is correct for other Windows versions.

The layout of version 2 is missing as well as the layout of other structures used by version 2.  All missing structures can be found at the link above.

### Steps to reproduce (optional)

_No response_

### Expected behavior (optional)

_No response_

### Actual behavior (optional)

_No response_

### Environment (optional)

```shell

```

diversenok:
This pull request improves API Set resolution in the following ways:
 - Adds v2 (Win 7, Win 8) and v4 (Win 8.1) definitions and support for them during parsing (closes #2519) 
 - Replaces linear search with binary search in the name lookup
 - Adds module-aware resolving, meaning that the choice of the host DLL can now depend on which module imports it. This is primarily meant for `kernel32.dll`, which previously caused peview to display cyclic dependencies, both on import and export pages (since peview was using the default resolution instead of the correct per-module one).


phlib/util.c
```diff
@@ -8610,65 +8610,421 @@ NTSTATUS PhDelayExecution(
...
-                    valueStringRef.Length = apisetMapValue->ValueLength;
-                    valueStringRef.Buffer = PTR_ADD_OFFSET(apisetMap, apisetMapValue->ValueOffset);
+            minIndex = 0;
+            maxIndex = Schema->Count - 1;
+            hash = PhpApiSetHashString(ApiSetName->Buffer, hashedLength, Schema->HashFactor);
 
-                    return PhCreateString2(&valueStringRef);
+            do
+            {
+                // Perform a binary search for the API Set name hash
```
jxy-s, line 8705: Why not use `bsearch` and provide a compare callback?

phlib/util.c
```diff
@@ -8610,65 +8610,421 @@ NTSTATUS PhDelayExecution(
...
+                return STATUS_NOT_FOUND;
 
-        if (PhStartsWithStringRef(ApiSetName, &nameStringRef, TRUE))
-        {
-            for (ULONG ii = 0; ii < apisetMapEntry->ValueCount; ii++)
+            if (Schema->Count > MAXINT32)
+                return STATUS_INTEGER_OVERFLOW;
+
+            // The hash covers the length up to but not including the last hyphen
+            for (hashedLength = ApiSetName->Length; hashedLength >= sizeof(WCHAR); hashedLength -= sizeof(WCHAR))
```
jxy-s, line 8690: Can this use `PhFindLastCharInStringRef` instead?

dmex (changes requested):

phlib/util.c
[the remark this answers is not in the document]
diversenok: `hashedLength` is the length in bytes, not an index (and it's treated as such). Either way, adjusted the code to use `PhFindLastCharInStringRef`.

dmex (approved):

jxy-s (approved):

PR #2606, opened 2025-06-19
Add thread RPC usage column

diversenok:
This pull request adds a column that indicates whether a thread has used RPC. It works by checking if `TEB->ReservedForNtRpc` is set.


SystemInformer/thrdlist.c
```diff
@@ -1295,6 +1347,7 @@ BOOLEAN NTAPI PhpThreadTreeNewCallback(
                     SORT_FUNCTION(Created), // Timeline
                     SORT_FUNCTION(ApartmentType),
                     SORT_FUNCTION(ApartmentFlags),
+                    SORT_FUNCTION(HasRpc),
```
dmex, line 1350: This is going to disable sorting of all thread columns? The array index for the sort function is the same as the column index and `"SORT_FUNCTION(HasRpc)"` needs to be the last entry since `PH_THREAD_TREELIST_COLUMN_RPC` is the last column.

SystemInformer/thrdlist.c
```diff
@@ -1993,6 +2046,26 @@ BOOLEAN NTAPI PhpThreadTreeNewCallback(
                     getCellText->Text = PhGetStringRef(node->ApartmentFlagsText);
                 }
                 break;
+            case PH_THREAD_TREELIST_COLUMN_RPC:
```
dmex, line 2049: The case position needs to be last to match the index.

dmex (changes requested):

SystemInformer/thrdlist.c
diversenok: Ahh, yes, my bad. I initially inserted the column right after COM flags but then decided to move it to the end to avoid reordering, and ended up making it worse. Good that you caught it 👍

jxy-s (approved):

PR #2638, opened 2025-08-01
Fix kernel APC routine in non-paged code

kkent030315:
- Fixed IRQL annotations in APC kernel/rundown routines
- APC kernel routines cannot be paged out, so put them in the non-paged code section

I think this is not yet triggered any bugs in the wild because the chance of getting those routines paged out is very low. Or are there any specific reason to be that they put in the paged code?


jxy-s:
> APC kernel routines cannot be paged out, so put them in the non-paged code section

@kkent030315 I do not believe this is true, where are you getting this information? I believe all types of APCs would be safe to be in a page-able code segment. There is perhaps an edge case around special kernel APCs when they're used in the context of filesystem I/O and that filesystem I/O is related to paging I/O? But I'll point out that none of the APCs here are used in that type of context.

kkent030315:
You are very correct and it was indeed on me that I have a fork of SI and somewhat temporarily raising IRQL to DPC, that was the cause. APC kernel routines can be safe to be paged out. Long time off from kernel development... Anyways, closing the PR. Thanks for correcting me :)

PR #2652, opened 2025-08-13
Fix unloaded module names

diversenok:
It looks like the return type for `PhCopyStringZ` changed from `BOOLEAN` to `NTSTATUS`, which caused the unloaded modules dialog to stop displaying DLL names.


dmex:
These two and another one were copypasta They were fixed in the other branch but have't been merged yet .

Issue #2674, opened 2025-09-04
Error in filtering conditions when parsing dynamic relocations


zjgcjy:
### Brief description of your issue

Here is the code in `PhpFillDynamicRelocations`, when the symbol equals `IMAGE_DYNAMIC_RELOCATION_GUARD_INDIR_CONTROL_TRANSFER`. When parsing data in the loop, it is necessary to determine whether the data is filling or aligning (similar to basic relocation), but the condition here is incorrect, and the loop ends prematurely, resulting in missing data.

https://github.com/winsiderss/systeminformer/blob/75c29283a41aa5c0638f1babe6cf72a381aabbf6/phlib/mapimg.c#L4544-L4548

In basic relocation, we use `IMAGE_REL_BASED_ABSOLUTE` to check such condition, which is indeed the high 4 bits. But in this case, it should be implemented as same as in case `IMAGE_DYNAMIC_RELOCATION_GUARD_SWITCHTABLE_BRANCH`.

https://github.com/winsiderss/systeminformer/blob/75c29283a41aa5c0638f1babe6cf72a381aabbf6/phlib/mapimg.c#L4595-L4598

That is, only when the page offset is zero does the data end.

### Steps to reproduce (optional)

Use peview.exe to open the test case on Windows 11 system, dxgkrnl.sys 10.0.26100.4230. Download URL:  https://msdl.microsoft.com/download/symbols/dxgkrnl.sys/1395d3f84e8000/dxgkrnl.sys.

View the data in the Dynamic Relocations column and find the one which `index==#26014`, `RVA==0x9f980`, `Type==INDIRECT`, `Info==CFG`.
<img width="1766" height="961" alt="Image" src="https://github.com/user-attachments/assets/ce089852-3f30-4224-aa39-1f0ae29bdd4e" />

Open the file with any hex editor at file `offset==004e017c` to observe such results.

<img width="1035" height="588" alt="Image" src="https://github.com/user-attachments/assets/154b7647-bcdf-4c3a-8d02-5cb27478063d" />

Here the VirtualAddress is 09f000, relocationCount is 4, which are 4980, 09b0, 49e0, 0000. The first item is consistent with the parsing result of peview.exe. The fourth item is the end data, but peview.exe did not parse the second and third items because it incorrectly exited the loop when parsing the second item `09b0`.

### Expected behavior (optional)

The other two relocations should be parsed out.

### Actual behavior (optional)

Due to incorrect filtering conditions, two relocations are missing.

[...]

PR #2675, opened 2025-09-04
fix: wrong condition in dvrt import control

zjgcjy:
See https://github.com/winsiderss/systeminformer/issues/2674 for details

### DIFF
--- a/phlib/mapimg.c
+++ b/phlib/mapimg.c
@@ -4541,8 +4541,7 @@ VOID PhpFillDynamicRelocations(
             {
                 PH_IMAGE_DYNAMIC_RELOC_ENTRY entry;
 
-                if (relocations[i].PageRelativeOffset == 0 ||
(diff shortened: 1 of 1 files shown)


zjgcjy:

zjgcjy:
recheck

jxy-s (approved):
Confirmed in via dumpbin that the check for the type being zero was errant:
```
    Symbol VA: 0000000000000004 IMAGE_DYNAMIC_RELOCATION_GUARD_INDIR_CONTROL_TRANSFER
    Fixup RVAs:
        [00000000] = page 0009F000 rva 0009F980 type 4
        [00000001] = page 0009F000 rva 0009F9B0 type 0
        [00000002] = page 0009F000 rva 0009F9E0 type 4

```

I believe the original intent was checking that the type was not `IMAGE_REL_*_ABSOLUTE` on the indirect case. But that's clearly irrelevant in this case.

Thanks for the patch ❤️

PR #2679, opened 2025-09-05
phnt: Fix NtGetNextThread flags parameter comment

m417z:
Context: https://github.com/m417z/ntdoc/pull/24.

cc @diversenok


jxy-s:
Confirmed via SRE

PR #2722, opened 2025-10-12
Use the UNDNAME_NO_MS_KEYWORDS parameter in the UnDecorateSymbolNameW…

kmlch-git:
Use the UNDNAME_NO_MS_KEYWORDS parameter in the UnDecorateSymbolNameW_I and IDiaSymbol_get_undecoratedNameEx functions.

### DIFF
--- a/phlib/symprv.c
+++ b/phlib/symprv.c
@@ -2649,11 +2649,21 @@ PPH_STRING PhUndecorateSymbolName(
 
     PH_LOCK_SYMBOLS();
 
+    /** For the same function, using the difference between the parameters of UNDNAME_COMPLETE and UNDNAME_NO_MS_KEYWORDS, obviously, using UNDNAME_NO_MS_KEYWORDS is more practical.
+    The function is:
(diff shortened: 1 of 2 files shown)


dmex:
Why does this flag needed t be added?

kmlch-git:
> Why does this flag needed t be added?

Please see the comments before calling the code of UnDecorateSymbolNameW_I function.

kmlch-git:
Developers don't need these MS KEYWORDS when they view and copy the C++ declaration of import and export functions of DLL or EXE files.

jxy-s:
I disagree, in your example the calling convention is an important detail which is lost with this change.

If there is a desire to configure the symbol decoration, should that be an advanced option instead?

kmlch-git:
I didn't actually use these MS KEYWORDS.
You can consider adding an option in PeView to control whether to use UNDNAME_COMPLETE or UNDNAME_NO_MS_KEYWORDS.

kmlch-git:
On the TAB page of Imports and Exports in PeView, select multiple lines, copy all the columns or a column in these lines, paste them on the clipboard, and output them to a separate TXT file. This function is very practical, and I have realized it.

PR #2759, opened 2025-11-30
DynData-less process protection

DavidXanatos:
For many users it has been a significant inconvenience that the driver refuses to operate when no up-to-date DynData is available. The driver previously required DynData for its process-protection subsystem; without it, the driver would not cooperate with the application for security reasons.

This pull request improves that behaviour by making process protection function reliably even without DynData. Previously, DynData was needed to retroactively revoke handle permissions when KphApplyObProtections was invoked before KphpCreateProcessNotifyRoutine had completed KphVerifyProcessAndProtectIfAppropriate, which could allow unrestricted handles during early process creation.

With this change, KphApplyObProtections will now lock and wait for KphVerifyProcessAndProtectIfAppropriate to finish, indicated by the DecidedOnProtection flag. If no process is tracked or DecidedOnProtection is not set within one second, tracking will be initiated (if needed) and the new AccessedDuringCreation flag will be set.

If AccessedDuringCreation is set, KphStartProtectingProcess will still require DynData so that the process’s handle list can be audited. If the flag is not set, we know no unapproved handles exist, so the handle audit can be safely skipped, allowing the function to complete successfully without DynData.

**Other changes**
1. In addition to SecurelyCreated, a new flag SecurelyCreatedSvc has been added. It is set when the process was created by a PPL process signed with PsProtectedSignerWinTcb (presumably services.exe). This allows service instances to use the driver when necessary.

2. KphProcessIsLsass no longer depends on DynData. It now first checks whether the target process is a PPL process signed with PsProtectedSignerLsa. If that check passes, no further (DynData-dependent) validation is performed. If it fails and no DynData is available, the function fails.

**Note:** For testing, I have added several additional changes guarded by IS_KTE #ifdefs and marked this pull request as a draft. If the maintainers consider the approach acceptable, I will clean up the code before marking the request as ready for review.

### DIFF
--- a/KSystemInformer/informer_process.c
+++ b/KSystemInformer/informer_process.c
@@ -101,7 +101,7 @@ PKPH_PROCESS_CONTEXT KphpPerformProcessTracking(
     if (!CreateInfo)
     {
 #ifdef IS_KTE
-        DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphpPerformProcessTracking Exiting: %s (%d)\n", PsGetProcessImageFileName(Process), (ULONG)(UINT_PTR)PsGetProcessId(Process));
+        //DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphpPerformProcessTracking Exiting: %s (%d)\n", PsGetProcessImageFileName(Process), (ULONG)(UINT_PTR)PsGetProcessId(Process));
 #endif
 
         process = KphUntrackProcessContext(ProcessId);
@@ -127,7 +127,7 @@ PKPH_PROCESS_CONTEXT KphpPerformProcessTracking(
     NT_ASSERT(CreateInfo);
 
 #ifdef IS_KTE
-    DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphpPerformProcessTracking Creating: %s (%d)\n", PsGetProcessImageFileName(Process), (ULONG)(UINT_PTR)PsGetProcessId(Process));
+    //DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphpPerformProcessTracking Creating: %s (%d)\n", PsGetProcessImageFileName(Process), (ULONG)(UINT_PTR)PsGetProcessId(Process));
 #endif
 
     process = KphTrackProcessContext(Process);
--- a/KSystemInformer/protection.c
+++ b/KSystemInformer/protection.c
@@ -575,26 +575,30 @@ NTSTATUS KphStartProtectingProcess(
     Process->ProcessAllowedMask = ProcessAllowedMask;
     Process->ThreadAllowedMask = ThreadAllowedMask;
 
-    // if the process wasn't marked as accessed we are done
-    if (!Process->AccessedDuringCreation)
-    {
-#ifdef IS_KTE
-        DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphStartProtectingProcess: Skipping KphEnumerateProcessContexts, process wasn't accessed\n");
-#endif
-        status = STATUS_SUCCESS;
-        goto Exit;
-    }
-
-#ifdef IS_KTE
-    DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphStartProtectingProcess: Starting KphEnumerateProcessContexts, process was accessed durign creation\n");
-#endif
-
     dyn = KphReferenceDynData();
     if (!dyn)
     {
-        status = STATUS_NOINTERFACE;
+        // if the process wasn't marked as accessed we are done
+        if (!Process->AccessedDuringCreation)
+        {
+#ifdef IS_KTE
+            DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphStartProtectingProcess: Skipping KphEnumerateProcessContexts, no dyn data available and process wasn't accessed\n");
+#endif
+            status = STATUS_SUCCESS;
+        }
+        else
+        {
+#ifdef IS_KTE
+            DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphStartProtectingProcess: Fail protection process was accessed, and there are no dyn data\n");
+#endif
+            status = STATUS_NOINTERFACE;
+        }
         goto Exit;
     }
+
+#ifdef IS_KTE
+    DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphStartProtectingProcess: KphpEnumProcessContextsForProtection\n");
+#endif
 
     context.Dyn = dyn;
     context.Status = STATUS_SUCCESS;
@@ -758,7 +762,7 @@ VOID KphApplyObProtections(
             if (!process && !hasTerminated)
             {
 #ifdef IS_KTE
-                DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphApplyObProtections Tracking: %s (%d)\n", PsGetProcessImageFileName(Info->Object), (ULONG)(UINT_PTR)PsGetProcessId(Info->Object));
+                //DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphApplyObProtections Tracking: %s (%d)\n", PsGetProcessImageFileName(Info->Object), (ULONG)(UINT_PTR)PsGetProcessId(Info->Object));
 #endif
                 process = KphTrackProcessContext(Info->Object);
 
@@ -777,7 +781,7 @@ VOID KphApplyObProtections(
                         KphDereferenceObject(process);
 
 #ifdef IS_KTE
-                        DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphApplyObProtections UnTracking: %s (%d)\n", PsGetProcessImageFileName(Info->Object), (ULONG)(UINT_PTR)PsGetProcessId(Info->Object));
+                        //DbgPrintEx(DPFLTR_DEFAULT_ID, 0xFFFFFFFF, "BAM KphApplyObProtections UnTracking: %s (%d)\n", PsGetProcessImageFileName(Info->Object), (ULONG)(UINT_PTR)PsGetProcessId(Info->Object));
 #endif
                         process = KphUntrackProcessContext(PsGetProcessId(Info->Object));
                         if (process)


jxy-s:
@DavidXanatos - thanks for the draft PR. As you may have already seen, the [`kph-staging` branch](https://github.com/winsiderss/systeminformer/compare/master...kph-staging) has a number of changes that are pending release. [One of them](https://github.com/winsiderss/systeminformer/commit/4c072456d3a5daafe147c19140609bade294bddc) introduces a requirement for at least TCB signing for `SecurelyCreated`, and an [additional commit](https://github.com/winsiderss/systeminformer/commit/f2bb9fd6e9a7c72a1f4e91f46c715d7d2165ab87) restarts the client through TCB. I’m pointing this out because the current changes in `kph-staging` already complement the direction of your proposal. Your contribution is good company for those changes, but it should be targeting the `kph-staging` branch, not `master`.

The real value here is in leveraging properties of PPL signing in the LSA path, combined with an attempt to detect whether a handle could exist outside of the creation path prior to its completion. If those conditions are comfortably satisfied, a client could, in theory, achieve the required end state without dynamic data.

That said, we must acknowledge that even with this type of change, certain APIs will not function if dynamic data is not present. In fact, today we could remove the client-side check and notification that dynamic data is incompatible and still load the driver. The client would simply operate in a lower-trust state and continue communicating with the driver accordingly. And as your requested changes demonstrate, we would still ultimately fall back to code paths that require dynamic data anyway.

If the primary concern is reducing user-facing “nagging” when dynamic data is out of date, that can already be achieved today with zero driver changes. However, the lack of dynamic data and reduced trust state will still restrict the available API surface exposed to the client. In practice, we therefore end up in the same functional situation with or without your changes - not only because we must still fall back to dynamic-data-dependent paths, but also because much of the meaningful information exposed from the driver to the client requires dynamic data. While the dialog may be considered overly verbose, it does clearly communicate to the user that capabilities are limited.

I cannot accept these changes as-is. As you pointed out in your PR description, the `IS_KTE` paths must be removed. Additionally:
1. It introduces an explicit `KeWaitForSingleObject` in the `Ob` path, which must be avoided.
2. The changes rely on LSA running as protected with the appropriate signing level. While this is the default on modern Windows systems, users can and do disable it. In those cases, we fall back to the dynamic-data-required path anyway. It also assumes a single process will ever run with the LSA signing level. While that is probably true today, the ALPC port owner check provides a more concrete relationship, and we cannot predict future changes to Microsoft’s use of the LSA signing level. I do believe this specific aspect is worth consideration, but it deserves more thought.
3. We must account not only for the natural process-creation flow, but also for third-party software that may intercept or alter that flow and pass handles across a boundary we cannot consider trusted. The current changes are not sufficiently considerate of this. There is an attempt to suspend those flows (see point 1), but that introduces significant application compatibility risk, degraded system performance, and a worse user experience.
4. If this work is realized, the client will be interfacing with the driver without dynamic data. Any subsequent API paths that require dynamic data will not achieve the desired behavior, and the user must be made aware of this. The changes attempt to convey this by adjusting the state level in the title, but that is realistically communicating the wrong information to the user. More consideration is required in this area.

All that said, I do appreciate the work here and acknowledge that something can likely be done to allow the client to reach the required state without dynamic data in certain paths. There is a clear benefit to reducing reliance on dynamic data, since some driver APIs do not actually require it and could be exposed if this is achieved.

While I’m unable to accept these changes verbatim, I will consider adopting some of the underlying ideas into the `kph-staging` branch where appropriate. If so, I will make those changes myself. This pull request can be closed.

DavidXanatos:
The primary motivation behind this work is to allow SI to use as much of the driver’s functionality as possible when DynData is unavailable, rather than falling back to the very restricted Low/Minimum states. From a quick pass over the code, it appears that most of the API surface does not intrinsically require DynData. As far as I can tell, only the following paths, and in some cases only their specific information classes, truly depend on it: KphpCommsAlpcQueryInformation, KphpCommsEnumerateProcessHandles, KphpCommsQueryInformationObject (for KphObjectSection*Information and KphObjectEtwRegBasicInformation), KphpCommsQueryInformationProcess (for KphProcessImageSection), KphpCommsStripProtectedProcessMasks, KphpCommsAssignProcessSessionToken, KphpCommsAssignThreadSessionToken, KphpCommsQueryInformationThread (for KphThreadIoCounters), KphpCommsQuerySection.

And process protection also currently requires DynData. That’s why I focused on enabling a safe path for process protection without DynData, doing so would unlock a substantial amount of functionality for users who currently must wait for DynData updates after every Windows release.

Regarding KeWaitForSingleObject in the Ob path: I understand the concern about blocking in this context. My reasoning was that the path already acquires KphAcquireRWLockShared(&process->ProtectionLock), which can itself be contended while KphStartProtectingProcess enumerates processes and handles. The additional wait would occur only in the specific case where KphApplyObProtections is entered before tracking has started, not in the general case where processes are accessed normally. The wait is strictly bounded (1 ms slices up to a fixed upper limit), so it cannot result in an indefinite block.

If necessary, this mechanism could be refined further so that this bounded wait is only performed for processes originating from the SI application directory, making that a prerequisite for protection eligibility.

A related improvement would be to add an early file-path eligibility check before probing for a *.sig file in KphVerifyProcessAndProtectIfAppropriate. This would avoid unnecessary *.sig lookups for processes that cannot be protected and would improve the current behaviour regardless of the DynData issue.

The goal of the “locking KphApplyObProtections” approach was to ensure that no unexpected handles exist at the moment protection is decided. In environments where third-party software intercepts or alters the process-creation flow, those systems could still rely on the existing DynData-dependent fallback. For the majority of systems, however, process protection would work reliably without DynData.

On LSASS: if LSASS is not running as PPL and DynData is unavailable, my suggestion was simply to deny access to protected processes and accept that some information may not be displayable in that configuration. I agree that this should be communicated clearly to the user when it occurs.

I have not yet focused on the user-facing side; for now I only needed a simple indication of whether DynData is present for testing. A proper implementation should clearly indicate which features are unavailable when DynData is missing or when the driver is not loaded. It could also notify the user if LSASS is not running as PPL and recommend enabling PPL when DynData cannot be obtained, so users understand what configuration is required to access the full feature set. From a user’s perspective, it seems important to make as much functionality available as possible without forcing them to wait on DynData updates, provided this can be done safely.

Finally, regarding the branch: would you prefer that any future work in this area target the kph-staging branch directly, or would you prefer to avoid PRs related to process protection and associated usability improvements at this time? If you are open to it, I am happy to rework this against kph-staging and remove the experimental IS_KTE paths.

jxy-s:
> The primary motivation behind this work is to allow SI to use as much of the driver’s functionality as possible when DynData is unavailable, rather than falling back to the very restricted Low/Minimum states. From a quick pass over the code, it appears that most of the API surface does not intrinsically require DynData. As far as I can tell, only the following paths, and in some cases only their specific information classes, truly depend on it: KphpCommsAlpcQueryInformation, KphpCommsEnumerateProcessHandles, KphpCommsQueryInformationObject (for KphObjectSection*Information and KphObjectEtwRegBasicInformation), KphpCommsQueryInformationProcess (for KphProcessImageSection), KphpCommsStripProtectedProcessMasks, KphpCommsAssignProcessSessionToken, KphpCommsAssignThreadSessionToken, KphpCommsQueryInformationThread (for KphThreadIoCounters), KphpCommsQuerySection.

Respectfully, the phrase “only the following paths” is doing a lot of work here - the list is both lengthy and includes several high-value code paths, which in practice reinforces the point that dynamic data remains a meaningful dependency.

> And process protection also currently requires DynData. That’s why I focused on enabling a safe path for process protection without DynData, doing so would unlock a substantial amount of functionality for users who currently must wait for DynData updates after every Windows release.

I don’t believe this premise is particularly compelling in practice. We are generally very quick to pick up Windows changes and publish updated dynamic data - typically within a 12–24 hour window, barring any disruption to our ability to release. Given that reality, the additional complexity, attack surface, and maintenance burden introduced by these changes are not justified by what is, at most, a short delay.

> Regarding KeWaitForSingleObject in the Ob path: I understand the concern about blocking in this context. My reasoning was that the path already acquires KphAcquireRWLockShared(&process->ProtectionLock), which can itself be contended while KphStartProtectingProcess enumerates processes and handles. The additional wait would occur only in the specific case where KphApplyObProtections is entered before tracking has started, not in the general case where processes are accessed normally. The wait is strictly bounded (1 ms slices up to a fixed upper limit), so it cannot result in an indefinite block.

A shared lock is far less expensive and serves an entirely different purpose than what is being introduced here. An explicit `KeWaitForSingleObject`, especially coupled with a 100 ms polling interval for up to one second, is fundamentally inconsistent with good kernel design and has no place in a high-frequency, latency-sensitive path such as an `Ob` callback.

> If necessary, this mechanism could be refined further so that this bounded wait is only performed for processes originating from the SI application directory, making that a prerequisite for protection eligibility.

The wait mechanism itself serves no useful purpose here, and further constraining it to processes originating from the client only simply introduces additional complexity without addressing the underlying problem. At this point, the reasoning is effectively demonstrating that the mechanism is unnecessary to begin with.

> A related improvement would be to add an early file-path eligibility check before probing for a *.sig file in KphVerifyProcessAndProtectIfAppropriate. This would avoid unnecessary *.sig lookups for processes that cannot be protected and would improve the current behaviour regardless of the DynData issue.

As currently described, this appears only partially considered. Depending on how that eligibility check is implemented, it could introduce a new bypass or inconsistency in the verification chain.

> The goal of the “locking KphApplyObProtections” approach was to ensure that no unexpected handles exist at the moment protection is decided. In environments where third-party software intercepts or alters the process-creation flow, those systems could still rely on the existing DynData-dependent fallback. For the majority of systems, however, process protection would work reliably without DynData.

How, specifically, do you propose we detect environments where third-party software intercepts or alters the process-creation flow? That is an unbounded and non-deterministic problem space, and cannot be reliably reasoned about in a generic way. Respectfully, this type of conditional assumption needs to be excluded from the design discussion unless a concrete and defensible detection method is provided.

> On LSASS: if LSASS is not running as PPL and DynData is unavailable, my suggestion was simply to deny access to protected processes and accept that some information may not be displayable in that configuration. I agree that this should be communicated clearly to the user when it occurs.

I believe there is some conceptual confusion here. The identification (or state) of LSASS is not what enforces the restriction on access to protected processes - that behaviour is the result of an explicit, hard-coded deny path. These are separate concerns that should not be conflated.

> I have not yet focused on the user-facing side; for now I only needed a simple indication of whether DynData is present for testing. A proper implementation should clearly indicate which features are unavailable when DynData is missing or when the driver is not loaded. It could also notify the user if LSASS is not running as PPL and recommend enabling PPL when DynData cannot be obtained, so users understand what configuration is required to access the full feature set. From a user’s perspective, it seems important to make as much functionality available as possible without forcing them to wait on DynData updates, provided this can be done safely.

Agreed - this area needs more consideration, as I noted in my previous message. Your reasoning here is generally sound.

> Finally, regarding the branch: would you prefer that any future work in this area target the kph-staging branch directly, or would you prefer to avoid PRs related to process protection and associated usability improvements at this time? If you are open to it, I am happy to rework this against kph-staging and remove the experimental IS_KTE paths.

As per my previous message, you do not need to contribute in this area. I will handle any changes myself. Respectfully, the implementation in this PR is fraught with problems and unresolved edge cases, and the accompanying commentary does not instill sufficient confidence in the underlying approach.

To be clear, as previously stated: “I will consider adopting some of the underlying ideas into the kph-staging branch where appropriate. If so, I will make those changes myself.”

DavidXanatos:
Thanks for taking the time to review the proposal and respond in detail. I hope some of the ideas can serve as inspiration for future improvements in this area.

DavidXanatos:
@jxy-s I believe there is a straightforward way to avoid introducing a wait in the Ob callback path. If the Ob callback is invoked before the process-creation callback, process tracking and protection could be initiated directly from the Ob callback instead.

Implementing this would require a somewhat larger change: specifically, moving KphAcquireRWLockExclusive(&Process->ProtectionLock) from KphStartProtectingProcess into KphVerifyProcessAndProtectIfAppropriate. Combined with restricting KphVerifyFile to run only when the executable path matches the driver’s parent application directory, this should provide a lightweight and robust solution without introducing any blocking waits.

Would this be an approach you would be willing to consider for a pull request?

jxy-s:
> @jxy-s I believe there is a straightforward way to avoid introducing a wait in the Ob callback path. If the Ob callback is invoked before the process-creation callback, process tracking and protection could be initiated directly from the Ob callback instead.
> 
> Implementing this would require a somewhat larger change: specifically, moving KphAcquireRWLockExclusive(&Process->ProtectionLock) from KphStartProtectingProcess into KphVerifyProcessAndProtectIfAppropriate. Combined with restricting KphVerifyFile to run only when the executable path matches the driver’s parent application directory, this should provide a lightweight and robust solution without introducing any blocking waits.
> 
> Would this be an approach you would be willing to consider for a pull request?

No.

PR #2778, opened 2025-12-15
phnt: Fix ThreadQuerySetWin32StartAddress comment

m417z:
According to @diversenok at https://github.com/m417z/ntdoc/pull/29#discussion_r2619425369:

> [T]he code for handling [`ThreadQuerySetWin32StartAddress`] in `NtSetInformationThread` looks like this:
> ```c
> case ThreadQuerySetWin32StartAddress:
>     return STATUS_INVALID_PARAMETER;
> ```


dmex:
Which OS version was tested? This infoclass was usable with NtSetProcessInformation on most earlier versions of Windows:

https://github.com/reactos/reactos/blob/ccfb061a2b1285c77d33eb154b5fbf8ebd069f1e/dll/win32/kernel32/client/proc.c#L457-L460

m417z:
From a quick check: it's usable in XP, returns `STATUS_INVALID_PARAMETER` in Win7 and in 10.0.26100.

PR #2847, opened 2026-03-11
Fix page/IO priority in Threads tab

ge0rdi:
Zero is valid value for page/IO priority. It was previously skipped due to wrong condition. Now these checks are the same as corresponding ones in `proctree.c`.


dmex:
✅ image

PR #2883, opened 2026-04-17
phnt: Revert accidental changes, fix typos

m417z:


dmex:
These 3 files in your PR are exact, 1:1 verbatim copies from the upstream Windows SDK, and any local changes will be overwritten the next time they're updated by the SDK.

You should report the issues upstream either https://github.com/MicrosoftDocs/ or https://developercommunity.visualstudio.com/VisualStudio/report

dmex (approved):

PR #2886, opened 2026-04-18
ntfill.h: annotate dynamically resolved routine names

m417z:
Added "// routine:" comments to `_Function_class_` markers for function typedefs that are resolved at runtime via `MmGetSystemRoutineAddress` or `RtlFindExportedRoutineByName`. Can be useful for readers and for docs generation.

Also, some of the pre-existing comments were incorrect (`CI_CHECK_SIGNED_FILE_EX` and `CI_VERIFY_HASH_IN_CATALOG_EX`).

I placed it at the same line as `_Function_class_` as I thought it'll make it less likely to go out of sync or be forgotten, but it results in longer lines. Let me know if you have a better idea. Perhaps an empty marco:
```cpp
#define _Routine_name_(name)
```

Usage (instead of the comment):
```cpp
_Function_class_(SE_UNREGISTER_IMAGE_VERIFICATION_CALLBACK)
_Routine_name_(SeUnregisterImageVerificationCallback)
```


KSystemInformer/include/ntfill.h
```diff
@@ -152,7 +152,7 @@ extern POBJECT_TYPE *IoDriverObjectType;
 extern POBJECT_TYPE *IoDeviceObjectType;
 
 typedef
-_Function_class_(IO_CHECK_FILE_OBJECT_OPENED_AS_COPY_SOURCE)
+_Function_class_(IO_CHECK_FILE_OBJECT_OPENED_AS_COPY_SOURCE) // routine: IoCheckFileObjectOpenedAsCopySource
```
jxy-s: Could we move these to the `typedef` please? The kernel code really tries to stick to an 80 character limit whenever possible, thanks!
```suggestion
typedef  // routine: IoCheckFileObjectOpenedAsCopySource
_Function_class_(IO_CHECK_FILE_OBJECT_OPENED_AS_COPY_SOURCE)
```

I did see your note about `#define _Routine_name_(name)` - but I don't think we should conflate SAL annotations with custom ones. Moving to the `typedef` would satisfy the width requirement in most (all?) cases.

KSystemInformer/include/ntfill.h
```diff
@@ -1738,9 +1736,9 @@ CI_CHECK_SIGNED_FILE_EX(
 typedef CI_CHECK_SIGNED_FILE_EX* PCI_CHECK_SIGNED_FILE_EX;
 
 // rev
-// CiVerifyHashInCatalog (pre 6.1.7601.18519)
+// pre 6.1.7601.18519
 typedef
-_Function_class_(CI_VERIFY_HASH_IN_CATALOG)
+_Function_class_(CI_VERIFY_HASH_IN_CATALOG) // routine: CiVerifyHashInCatalog
```
jxy-s: I'm curious if we could move the "pre 6.1.7601.18519" to the "rev" line or if that would be too complicated for your parser?

```suggestion
// rev (pre 6.1.7601.18519)
// routine: CiVerifyHashInCatalog
typedef
_Function_class_(CI_VERIFY_HASH_IN_CATALOG) 
```

jxy-s (changes requested):

KSystemInformer/include/ntfill.h
```diff
@@ -1718,8 +1716,7 @@ CI_CHECK_SIGNED_FILE(
 typedef CI_CHECK_SIGNED_FILE* PCI_CHECK_SIGNED_FILE;
 
 // rev
-// CiCheckSignedFile
-typedef
+typedef // routine: CiCheckSignedFileEx
```
jxy-s: Oh... a note here... sorry I missed this is my last review. 

This is actually the same routine name exported by `CI.dll`... they broke the signature between `6.1.7601.18519` and later versions. So this should technically be `CiCheckSignedFile` like the one before it. I personally put an `_EX` on the `typedef` so the symbols did not conflict and an older version of KSystemInformer would cast to the right signature based on what `CI.dll` version it was dealing with - that's not longer the case since we reliably use `CiValidateFileObject` instead of these.

The same goes for `CiVerifyHashInCatalogEx` -> `CiVerifyHashInCatalog` below. Since you're aiming for documentation generation you probably want some linkage between the two indicating that there is a signature change?

m417z: I was sure `CI_CHECK_SIGNED_FILE_EX` maps to `CiVerifyHashInCatalogEx` since it's MS API convention. Not sure `CI_CHECK_SIGNED_FILE_EX` is the best name, but for the scope of this PR: Yes, for documentation, I'd like both to map to `CiVerifyHashInCatalog`. For the comments, how about specifying the relevant Windows versions? e.g.:
```cpp
typedef // routine: CiCheckSignedFile (before 6.1.7601.18519)
...
CI_CHECK_SIGNED_FILE ...
```
```cpp
typedef // routine: CiCheckSignedFile (6.1.7601.18519 and later)
...
CI_CHECK_SIGNED_FILE_EX ...
```

jxy-s: Yeah that's reasonable 👍. Also sorry about the `_EX` that's certainly poor naming on my part. If you're okay with it - as part of this PR let's rename the `typedef` to avoid that confusion in the future, maybe `_LEGACY` instead to avoid the `_EX` convention, like this?

```
typedef // routine: CiCheckSignedFile (before 6.1.7601.18519)
_Function_class_(CI_CHECK_SIGNED_FILE_LEGACY)
```
```
typedef // routine: CiCheckSignedFile (6.1.7601.18519 and later)
_Function_class_(CI_CHECK_SIGNED_FILE)
```

jxy-s (approved):

PR #2897, opened 2026-05-04
memsrcht: parallel memory string scan with performance optimizations

Zack-src:
Rewrites the memory strings scan engine in `memsrcht.c` for performance and memory efficiency. All changes are self-contained to this file.

### Parallel scan (commit 1)

Pre-enumerate all committed, accessible memory regions that match the selected type mask (Private/Image/Mapped), then distribute them across N worker threads (N = logical CPU count, capped at 64, configurable via Settings). Each worker independently reads and scans its assigned regions using `PhpMemoryStringSearchTreeNextBufferParallel`. Read buffer increased from 16 MiB to 32 MiB to reduce `NtReadVirtualMemory` call frequency. `StringsCount` is incremented atomically via `InterlockedIncrement` for thread safety.

### Slab allocator, string arena, batch locking, lazy formatting (commit 2)

Replace per-string heap allocations with two pooled allocators:
- **Node slab**: 256-node fixed-size blocks. Eliminates ~29M individual `PhAllocateZero` calls during a typical Chrome scan.
- **String arena**: 16 MB `WCHAR` buffers per worker thread. Eliminates ~29M `PhCreateString2` calls; string data is packed contiguously.

`PPH_STRING String` in `PH_MEMSTRINGS_NODE` is replaced by `PH_STRINGREF StringRef` pointing into the arena. String lifetime is tied to the arena, freed atomically in `PhpDeleteMemoryStringsTree`.

Workers accumulate 1024 nodes locally before acquiring `SearchResultsLock`, reducing lock contention from ~29M to ~28K acquisitions per scan.

Five pre-formatted WCHAR arrays in `PH_MEMSTRINGS_NODE` are replaced by five `PPH_STRING` lazy fields. Display strings are allocated only when a row becomes visible in `GetCellText`, not eagerly for every node.

Thread count is user-configurable via a new Settings menu item (0 = auto).

### Bulk node insertion and skip unnecessary qsort (commit 3)

`PhpAddPendingMemoryStringsNodes` now uses `PhResizeList` to pre-allocate `NodeList` capacity, skips redundant `memset` calls (nodes are pre-zeroed by the slab), and appends all pending nodes in one `PhAddItemsList` call instead of N individual `PhAddItemList` calls. Filter evaluation is skipped when no search text is active.

`qsort_s` in `TreeNewGetChildren` is skipped when the sort column is Index in ascending order - nodes are produced in monotonically increasing index order, so the list is already sorted.

### Remove TextCache to halve per-node memory (commit 4)

`PH_STRINGREF TextCache[8]` (128 bytes per node) is removed from `PH_MEMSTRINGS_NODE`. The five lazy `PPH_STRING` fields serve as the per-node cache; `GetCellText` is a pointer lookup after first access. `TN_CACHE` is dropped accordingly.

Node size: 280 bytes > 152 bytes (−45%).
Theoretical memory for 31M nodes: 8.3 GB > 4.5 GB (−3.8 GB).
Observed peak process memory on a Chrome scan: ~20 GB > ~6.5 GB.

Fix: `StringArenaList` (`PPH_LIST` object) was never released - `PhClearReference` added after `PhpFreeMemoryStringsArenas` in `PhpDeleteMemoryStringsTree`.

### DIFF
--- a/SystemInformer/memsrcht.c
+++ b/SystemInformer/memsrcht.c
@@ -125,8 +125,9 @@ typedef struct _PH_MEMSTRINGS_NODE
     PPH_STRING AddressText;
     PPH_STRING LengthText;
     PPH_STRING ProtectionText;
-
-    PH_STRINGREF TextCache[PH_MEMSTRINGS_TREE_COLUMN_ITEM_MAXIMUM];
+    // No TextCache array: PPH_STRING fields serve as the per-node cache.
+    // GetCellText does not set TN_CACHE so TreeNew calls us per visible cell,
+    // but each call is a pointer lookup once the string is allocated.
 } PH_MEMSTRINGS_NODE, *PPH_MEMSTRINGS_NODE;
 
 #define MEMSTRINGS_SLAB_CAPACITY 256
@@ -890,6 +891,7 @@ VOID PhpDeleteMemoryStringsTree(
     PhpDeleteMemoryStringsNodeList(Context->NodeList);
     PhpFreeMemoryStringsSlabs(Context);
     PhpFreeMemoryStringsArenas(Context);
+    PhClearReference(&Context->StringArenaList);
 
     PhDeleteTreeNewFilterSupport(&Context->FilterSupport);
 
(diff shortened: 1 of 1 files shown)


dmex (approved):

jxy-s (approved):
Nice, thanks!

PR #2948, opened 2026-06-22
Implement column type "Short user name"

### LINKED ISSUE #2845: Hide "uninteresting" domain names in UI
### Description of the feature, modification, idea or suggestion

Basically, I would like to avoid seeing (pseudo-)domain names like `NT AUTHORITY` or _`MY_HOSTNAME`_ or _`MY_DOMAIN`_ prefixed to any account name in the `User name` column.

### Proposed implementation details (optional)

I think this means you'd want `PhGetSidFullName` to hide the domain name for the following cases:
- For `PhSeLocalSystemSid`
- When the domain SID matches that returned by `LsaQueryInformationPolicy(PhGetLookupPolicyHandle(), PolicyAccountDomainInformation, ...)`

YeDemirkiran:
This PR implements the feature requested at #2845

## Description
This PR introduces a new process tree column called "Short user name" that strips out local or machine domain prefixes (e.g., "NT AUTHORITY\"), making the process list cleaner.

## Implementation
- Added `PHPRTLC_SHORT_USERNAME` enum (and its sort functions) and ShortUsername column type 
- Find the address of last backslash inside UserName, create a new string from it and pass it to ShortUsername
- Use the variable inside the TreeNewGetCellText case in the PhpProcessTreeNewCallback function

## Screenshots
image
image

<hr>

Closes #2845


YeDemirkiran:

SystemInformer/proctree.c
dmex, line 4720: Ref

dmex, line 4730: Ref

SystemInformer/proctree.c
```diff
@@ -4698,6 +4711,22 @@ BOOLEAN NTAPI PhpProcessTreeNewCallback(
                     }
                 }
                 break;
+            case PHPRTLC_SHORT_USERNAME:
+                {
+                    if (!processItem->ShortUserName && processItem->UserName)
```
dmex: You've added everything correctly but there's a nuance with how nodes are only created when visible. If you try sorting the column it won't create all the strings and only sorts the visible items.

The other columns have functions like PhpUpdateProcessNodeServices,  PhpUpdateProcessNodeMitigationPolicies etc... and you need to add a similar function and flag for this column otherwise the sorting and other stuff won't work properly.  I can fix it up for you if you want or you can add it?

dmex (changes requested):

SystemInformer/proctree.c
YeDemirkiran: Hey, thanks for the review. Yes, I can add the function. Working on it right now. I'm new to the codebase so I might miss some important stuff. Is the PhpUpdateProcess function and flag the only necessary addition in this case?

Also, Copilot generated a fix below. If it's good, we can use it.

dmex (approved):

SystemInformer/proctree.c
```diff
@@ -1910,6 +1911,27 @@ static VOID PhpUpdateProcessNodeServices(
...
+        PhClearReference(&ProcessNode->ShortUsernameText);
+
+        if (ProcessNode->ProcessItem->UserName)
+        {
+            wchar_t* backslash = wcsrchr(ProcessNode->ProcessItem->UserName->Buffer, L'\\');
+            if (backslash)
+                ProcessNode->ShortUsernameText = PhCreateString(backslash + 1);
+            else
+                ProcessNode->ShortUsernameText = PhCreateString(ProcessNode->ProcessItem->UserName->Buffer);
+        }
```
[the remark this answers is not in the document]
YeDemirkiran: Will fix it

dmex (approved):

dmex:
🎉

PR #2970, opened 2026-07-05
Show empty status message when Windows Defender Firewall is disabled

NotEnsoul:
### Overview
If Windows Defender Firewall is disabled (or its service is stopped), the Firewall tab in System Informer displays a completely empty list. This pull request adds a detection helper and sets a descriptive empty status message so the user is informed that the firewall is disabled.

---

### Implementation Details
* **Empty Text Definition**: Added `FwTreeDisabledText` in [fwtab.c]:
  ```c
  CONST PH_STRINGREF FwTreeDisabledText = PH_STRINGREF_INIT(L"The Windows Defender Firewall is disabled.");
  ```
* **Firewall Status Detection**: Implemented `EtFwIsFirewallEnabled()` in [fwtab.c]:
  * Queries the status of the `mpssvc` (Windows Defender Firewall) service. If the service is not running, the firewall is considered disabled.
  * Queries the registry value `EnableFirewall` under `SharedAccess\Parameters\FirewallPolicy` for the `DomainProfile`, `StandardProfile` (Private), and `PublicProfile`. If all three profiles are explicitly set to `0` (disabled), it returns `FALSE`.
* **Empty Text Update**: If the event tracing session is running successfully (`EtFwEnabled` is `TRUE`), but the firewall is disabled, the tree view's empty status text is set to `FwTreeDisabledText`.


dmex (approved):

plugins/ExtendedTools/fwtab.c
```diff
@@ -80,6 +81,65 @@ PPH_POINTER_LIST EtFwNodeStateList = NULL;
...
+    VOID
+    )
+{
+    BOOLEAN enabled = TRUE;
+    SC_HANDLE serviceHandle;
+    SERVICE_STATUS_PROCESS statusProcess;
+
+    if (NT_SUCCESS(PhOpenService(&serviceHandle, SERVICE_QUERY_STATUS, L"mpssvc")))
+    {
+        if (NT_SUCCESS(PhQueryServiceStatus(serviceHandle, &statusProcess)))
```
dmex: The SCM can block for a long time and we already have the service state. In this case using the cached state should be used instead of querying the system:

```
PPH_SERVICE_ITEM serviceItem;

if (serviceItem = PhReferenceServiceItemZ(L"mpssvc"))
{
    if (serviceItem->State == SERVICE_RUNNING)
    {
        enabled = TRUE;
    }

    PhDereferenceObject(serviceItem);
}
```

dmex (changes requested):

NotEnsoul:
i noticed that even though windows firewall is disabled(non-service) we can still see WSH embedded rules in WPF for a short period of time

obraz

obraz

plugins/ExtendedTools/fwtab.c
NotEnsoul: applied

dmex (approved):

dmex:
FWGetGlobalConfig + FW_PROFILE_CONFIG_ENABLE_FW = [firewallapi](https://github.com/user-attachments/files/29894872/firewallapi.txt)

> i noticed that even though windows firewall is disabled(non-service) we can still see WSH embedded rules in WPF for a short period of time

This is because mpssvc is only a client of the `Base Filtering Engine` which is the actual firewall and disabling mpssvc doesn't do anything but disable defender.

dmex:
> This is because mpssvc is only a client

Here's a diagram:

image

System Informer registers callbacks with the Base Filtering Engine and is able to show events from both Windows Firewall and third-party firewall software (zonealarm and various antivirus products etc...).

The changes in this PR for checking the Windows Defender mpssvc service and Windows Defender Firewall policies to determine when the firewall is disabled are insufficient and not correct because they are both clients of the Base Filtering Engine and BFE remains active when they are disabled.

> Services\\SharedAccess\\Parameters\\FirewallPolicy

BFE has an entirely different set of firewall rules and policies. You can view them with tools like https://github.com/zodiacon/WFPExplorer

NotEnsoul:
we are unable to track network connectivity whenever BFE is disabled and receive 1753 error so we could maybe add a message for this specific status code instead of a "There are no more endpoints from the endpoint mapping program" ???(this status message is translated from my language)

NotEnsoul:
i was thinking of what to do and tried some things but i don't think proper disabled firewall message can exist due to preexisting wfp firewall rules, the only thing we can do to improve user experience is perhaps leaving a message that the tracking is still working even though nothing may appear for x minutes while on the firewall page

PR #2971, opened 2026-07-05
Fix GPU/NPU process resident memory columns and add committed memory columns

NotEnsoul:
## Description
This change resolves the issue where the process-level GPU memory columns displayed virtual committed memory sizes rather than the actual resident VRAM usage, causing a discrepancy with the system-wide dedicated usage. It also implements new columns to track committed GPU and NPU memory explicitly.

### Key Changes:
1. **GPU/NPU Memory Correction (Resident VRAM):**
   - Mapped the process-level dedicated and shared bytes columns to the resident memory performance counter indices (`Local Usage` / index 2 and `Non Local Usage` / index 3).
   - Refactored the query fallback path (when D3D perf counters are disabled) in `gpumon.c` and `npumon.c` to use `D3DKMT_QUERYSTATISTICS_PROCESS_SEGMENT_GROUP` to obtain actual resident `Usage` instead of accumulating `BytesCommitted`.

2. **Added Committed Memory Columns:**
   - Created four new process columns:
     - `GPU dedicated bytes (committed)`
     - `GPU shared bytes (committed)`
     - `NPU dedicated bytes (committed)`
     - `NPU shared bytes (committed)`
   - Linked these columns to WDDM performance counters (`Dedicated Usage` / index 4 and `Shared Usage` / index 5) and implemented segment-by-segment `BytesCommitted` loops in the fallback paths.

3. **UI & Formatting Support:**
   - Updated process tree definitions, column formatting, sorting, and header-totals calculations in `treeext.c` to support the new columns.


dmex (approved):

dmex:
I need to test these changes on LTSB because of differences with Windows 10 and 11. I also need to know which version of Windows were you using to debug/test these changes?

NotEnsoul:
> I need to test these changes on LTSB because of differences with Windows 10 and 11. I also need to know which version of Windows were you using to debug/test these changes?

I have tested on latest W11 insider 26300.8758

What I noticed about resident vram tracking is that it starts falling off once total vram usage goes really high ~15gb at that point of time resident shows about 9gb used, but we are going through per process information for vram usage so maybe it is driver/framework api things

dmex:
> latest W11 insider 26300.8758

Microsoft deprecated D3DKMT statistics for PerfCounterV2 on Windows 10 a few years ago, then last year decided to revert PerfCounterV2 back to D3DKMT

> What I noticed about resident vram tracking is that it starts falling off once total vram usage goes really high ~15gb at that point of time resident shows about 9gb used, but we are going through per process information for vram usage so maybe it is driver/framework api things

D3DKMT_QUERYSTATISTICS_SEGMENT_INFORMATION should contain information about memory owned by the driver and kernel that isn't attributed to any process. If you enumerate every segment, filter to dedicated segments via the Aperture flag and SegmentProperties bitfield (SegmentType, SegmentGroup, SystemMemory, FullyCPUVisible), and sum BytesResident... This should report the 15 GB, now subtract the summed per-process total and you should have the missing 6 GB which is "unattributed / driver-reserved".

D3DKMT_QUERYSTATISTICS_SEGMENT_USAGE / SEGMENT_GROUP_USAGE are recent additions (also the ADAPTER2/SEGMENT2/PROCESS_SEGMENT_GROUP2 variants) and return D3DKMT_QUERYSTATISTICS_MEMORY_USAGE which doesn't require enumerating every segment and a very fast method of calculating the local/non-local usage but it requires Windows 11 - these might be required for the correct totals.

PR #2974, opened 2026-07-06
devices: Fix registry key navigation in Devices> Regedit

NotEnsoul:
## Description
This PR resolves two issues that prevent the Device Tree and PnP Services registry actions (right-click -> Key -> Hardware/Software/User/Config) from correctly opening and navigating to target keys in Registry Editor (`regedit`).

---

### Root Cause
1. **Hardware Key Logic Bug**: In `phlib/util.c`, `PH_DEVKEY_HARDWARE` is defined as `0x00000000`. The code used `else if (FlagOn(Flags, PH_DEVKEY_HARDWARE))` to check for this flag. Because `FlagOn` checks `Flags & PH_DEVKEY_HARDWARE` (effectively `Flags & 0`), it always returned `FALSE`, making the hardware key retrieval block dead code.
2. **NT vs Win32 Registry Path Format Mismatch**: The device tree plugin retrieved the native NT registry path (e.g., `\Registry\Machine\System\...`) from the kernel key handle and passed it directly to `PhShellOpenKey2`. Because `regedit` expects Win32 registry roots (e.g., `HKEY_LOCAL_MACHINE\...`), it opened but failed to navigate to the key, leaving the user on the default or last opened key.

### Fix
1. Changed the check for `PH_DEVKEY_HARDWARE` in `PhDevOpenObjectKey` to a direct equality check (`Flags == PH_DEVKEY_HARDWARE`).
2. Updated both `plugins/HardwareDevices/main.c` and `plugins/ExtendedServices/svcpnp.c` to update `bestObjectName` in-place using `PhMoveReference` combined with `PhFormatNativeKeyName`. This converts the NT path format to a Win32 format seamlessly while safely reusing the existing `PhDereferenceObject` lifetime management logic.

### Steps to Verify
1. Run `SystemInformer.exe` (with PnP plugins enabled).
2. Open the **Devices** tab.
3. Right-click on any device (e.g., `AMD GPIO Controller`) and select **Key** -> **Hardware** (or **Software**).
4. Verify that Regedit opens and successfully highlights/selects the correct registry key.

### DIFF
--- a/phlib/util.c
+++ b/phlib/util.c
@@ -11764,7 +11764,7 @@ NTSTATUS PhDevOpenObjectKey(
                 }
             }
         }
-        else if (Flags == PH_DEVKEY_HARDWARE)
+        else if (FlagOn(Flags, PH_DEVKEY_HARDWARE))
         {
             // HARDWARE Key: \Registry\Machine\System\CurrentControlSet\Enum\{InstanceID}\Device Parameters
             static const PH_STRINGREF base = PH_STRINGREF_INIT(L"System\\CurrentControlSet\\Enum\\");
--- a/phlib/util.c
+++ b/phlib/util.c
@@ -11764,7 +11764,7 @@ NTSTATUS PhDevOpenObjectKey(
                 }
             }
         }
-        else if (FlagOn(Flags, PH_DEVKEY_HARDWARE))
+        else if (Flags == PH_DEVKEY_HARDWARE)
         {
             // HARDWARE Key: \Registry\Machine\System\CurrentControlSet\Enum\{InstanceID}\Device Parameters
             static const PH_STRINGREF base = PH_STRINGREF_INIT(L"System\\CurrentControlSet\\Enum\\");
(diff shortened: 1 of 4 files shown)


NotEnsoul:
or change flag

phlib/util.c
```diff
@@ -11764,7 +11764,7 @@ NTSTATUS PhDevOpenObjectKey(
                 }
             }
         }
-        else if (FlagOn(Flags, PH_DEVKEY_HARDWARE))
+        else if (Flags == PH_DEVKEY_HARDWARE)
```
jxy-s: I think we should probably just fix the flag values.
```diff
-#define PH_DEVKEY_HARDWARE        (0x00000000)
-#define PH_DEVKEY_SOFTWARE        (0x00000001)
-#define PH_DEVKEY_USER            (0x00000100)
-#define PH_DEVKEY_CONFIG          (0x00000200)
+#define PH_DEVKEY_HARDWARE        (0x00000001)
+#define PH_DEVKEY_SOFTWARE        (0x00000002)
+#define PH_DEVKEY_USER            (0x00000004)
+#define PH_DEVKEY_CONFIG          (0x00000008)
```

NotEnsoul:
applied

dmex (approved):

PR #2989, opened 2026-07-12
do not steal focus on hidden startup

### LINKED ISSUE #2242: Don't change Focus
### Brief description of your issue

Wish:
If Set option: [x] Start hidden -
don't change Focus from current window (where start SystemInformer.exe)

### Steps to reproduce (optional)

Real example problem:
I many time manually start/restart SystemInformer.exe from Far Manager
my SystemInformer.exe.settings.xml have option  [x] Start hidden
and after start program every time I lost focus from Far Manager window.

### Expected behavior (optional)

May be need add logic:

    main()
    Old_hwnd: HWND;
    Old_hwnd:= GetForegroundWindow();

    ...
    if option.hide == true
       {
        SetForegroundWindow(Old_hwnd);
       }

where "option.hide" it's:
ListView_GetCheckState(listViewHandle, PHP_OPTIONS_INDEX_START_HIDDEN) == BST_CHECKED
from options.c

### Actual behavior (optional)

_No response_

### Environment (optional)

```shell
SI 3.1.24290.0
WIN32_NT 10.0.19045.3448 (22H2)
```

Zeroes1:
Fix: Prevent main window from stealing focus on hidden startup
Fix: https://github.com/winsiderss/systeminformer/issues/2242 (has not been fixed before)

Problem: On Windows OS, when System Informer is configured to start hidden (tray icon only), the creation of the first top-level window ("Startup window") unconditionally steals foreground focus from the current application — even though ShowWindow is called with SW_HIDE. This is a Windows OS behavior: the first top-level window of a new process is automatically activated by the window manager regardless of the show command.

Root cause: CreateWindowEx at mainwnd.c:120 creates the main window without WS_EX_NOACTIVATE. Windows then implicitly activates this window because it is the first visible top-level window owned by the process. The subsequent ShowWindow(SW_HIDE) does not undo the activation.

Fix: Conditionally apply WS_EX_NOACTIVATE to CreateWindowEx when the window is known to start hidden. The condition mirrors the existing hidden-start logic in PhMwpShowWindow:

ShowHidden startup parameter OR SETTING_START_HIDDEN enabled
ShowVisible startup parameter is NOT set
Notification area (tray) icons are enabled (PhNfIconsEnabled())
When these criteria are met, the extended style WS_EX_NOACTIVATE prevents Windows from auto-activating the window during creation. When not met, no style flag is set, preserving normal activation behavior without visual flicker.

Testing:
I compiled with the fix and checked it manually.
a) Hidden start: window no longer steals focus from the foreground app
b) No Hidden start: as before, no problems


Zeroes1:

dmex:
> Problem: On Windows OS, when System Informer is configured to start hidden (tray icon only), the creation of the first top-level window ("Startup window") unconditionally steals foreground focus from the current application — even though ShowWindow is called with SW_HIDE. 

The main window is always created invisible by default (without the WS_VISIBLE flag):
https://github.com/winsiderss/systeminformer/blob/5bb68c3a978533680e782bff60abe9d979ca6b64/SystemInformer/mainwnd.c#L120-L132

ShowWindow/SetForegroundWindow is also never called with SW_HIDE:
https://github.com/winsiderss/systeminformer/blob/5bb68c3a978533680e782bff60abe9d979ca6b64/SystemInformer/mainwnd.c#L494-L504

>  the creation of the first top-level window ("Startup window") unconditionally steals foreground focus from the current application

The window is invisible and doesn't change foreground.

> This is a Windows OS behavior: the first top-level window of a new process is automatically activated by the window manager regardless of the show command.

Windows doesn't change activation or foreground by default. The default behavior requires the application to change the foreground window and this is limited to the current application.

See remarks:
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow#remarks
Also comments:
https://learn.microsoft.com/en-us/windows/win32/winmsg/using-windows#creating-a-main-window
Also remarks:
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setforegroundwindow#remarks

Zeroes1:
ok, more deep analysis showed:

PhMwpSelectionChangedTabControl(INT_ERROR) at startup — activates the hidden main window through its child control. Even though the window was created without WS_VISIBLE and never shown, SetFocus on a descendant triggers WM_ACTIVATE up the ancestor chain, making the hidden window the foreground.

`SetFocus(page->WindowHandle)` at mainwnd.c:4737

fix:
```
                if (!((PhStartupParameters.ShowHidden || PhGetIntegerSetting(SETTING_START_HIDDEN)) && !PhStartupParameters.ShowVisible && PhNfIconsEnabled()) || IsWindowVisible(PhMainWndHandle))
                    SetFocus(page->WindowHandle);
```

This skips SetFocus only during hidden startup; SetFocus still fires when the window is visible (tab switching after showing from tray, or normal visible startup).

I checked it manually - yes, it helps fix the problem, but the original version also solved it :)

PR #2990, opened 2026-07-12
Services: Fall back to registry for protected configuration changes

NotEnsoul:
# Description

## Problem

Some protected services reject configuration changes through the Service Control Manager even after elevation. The service properties dialog previously reported the failure without providing another path for applying the requested configuration.

Local debug builds could also fail to connect to their elevated phsvc helper when an installed release helper owned the shared port, or when API-build signature validation rejected the unsigned debug executable.

## Solution

- Fall back to updating the service registry key when SCM configuration fails.
- Open protected service keys with backup/restore semantics under the elevated helper.
- Support service type, start type, error control, image path, load-order group, account name, display name, and delayed auto-start values.
- Do not use the registry fallback for password changes because service passwords are stored as LSA secrets.
- Return whether phsvc used the registry fallback so the UI can explain that Windows must be restarted.
- Allow an unelevated System Informer instance to apply the fallback through its temporary UAC-elevated phsvc helper.
- Give debug builds separate native and WOW64 phsvc port names and retain same-executable validation for unsigned local debug binaries.
- Preserve catalog hash errors before cleanup and correctly normalize Win32-facility HRESULT values discovered while testing the debug build.

## Notes

Registry fallback changes are not observed by the running Service Control Manager. The dialog informs the user that Windows must be restarted before the new service configuration takes effect.


SystemInformer/srvprp.c
```diff
@@ -813,84 +820,85 @@ INT_PTR CALLBACK PhpServiceGeneralDlgProc(
...
+                        }
 
-                            PhMarkNeedsConfigUpdateServiceItem(serviceItem);
+                        PhCloseServiceHandle(serviceHandle);
+                    }
 
-                            PhCloseServiceHandle(serviceHandle);
-                        }
-                        else
+                    if (!NT_SUCCESS(status))
```
jxy-s, line 865: Writing directly to the SCM-owned registry keys puts the SCM's internal state out of sync with what's persisted. This works most of the time, but it's not without consequences:

- Some services are intentionally configured without a stop capability to protect against rendering a machine unbootable. Bypassing the SCM bypasses that protection.
- Editing the service entry of a product with additional mitigations in place (AV/EDR, etc.) is likely to raise a detection with that product.

Given that, I'd like to see a few changes here:

1. Gate this behind an advanced option rather then enabling it unconditionally.
2. Only attempt the fallback on explicit status, not any SCM failure (e.g. `STATUS_ACCESS_DENIED`). Falling back immediately whenever the blessed SCM path fails is unwise.
3. Prompt the user _before_ the registry edit (Yes/No), explaining that this will write to the registry directly and that a reboot is required before the change takes effect. I see the post-hoc restart message, but since the change won't be reflected in the UI immediately, users need to opt in and make the mental note up front - a notification after the fact is too late for them to reconsider.

dmex:
The PR is on hold temporarily due to pending changes to phsvc.

NotEnsoul:
obraz

SystemInformer/srvprp.c
NotEnsoul: applied in newest commit

jxy-s (approved):
LGTM, thanks! Leaving my approval, but let's wait on @dmex to unblock before merging.

PR #3001, opened 2026-07-19
ExtendedTools: Add summary GPU engine column

### LINKED ISSUE #2637: Add GPU Engine Column
### Description of the feature, modification, idea or suggestion

Currently when viewing the GPU %, it doesn't tell us *WHICH* GPU it is for. We currently have to go back and forth between SysInformer and TaskManager which provides which GPU that process is actually using.

### Proposed implementation details (optional)

Add the same GPU engine column that we can add to inform which GPU and what engine it is running.

NotEnsoul:
### Description

This PR implements a summary "GPU engine" column in the process tree (similar to the one in Windows Task Manager). It displays a summary of which GPU adapter and node are being used the most by each process.

### Details
- Lists `GPU <adapter_index> - <node_name>` (e.g. `GPU 0 - 3D`) or `GPU <adapter_index> - Node <local_node_index>` (e.g. `GPU 0 - Node 4` if no node description is available) for the node with the highest utilization.
- Only displays when utilization is equal to or greater than `0.01%` (utilization >= `0.0001f`), matching Task Manager behavior.
- Supports sorting by the maximum GPU node utilization.
- Fully supports child process GPU utilization propagation.
- Handles cell caching and invalidation correctly when process updates tick or when tree nodes expand.

---

Now actually resolves https://github.com/winsiderss/systeminformer/issues/2637 (i understood it wrong before!)


plugins/ExtendedTools/treeext.c
```diff
@@ -143,6 +143,17 @@ VOID EtProcessTreeNewInitializing(
...
+        EtpAddTreeNewColumn(
+            treeNewInfo,
+            ETPRTNC_GPUENGINE,
+            L"GPU engine",
+            80,
+            PH_ALIGN_LEFT,
+            DT_LEFT,
+            TRUE,
+            EtpProcessTreeNewSortFunction
+            );
```
jxy-s: This should probably go in the list above around line 108?

jxy-s (changes requested):
Question regarding placement of the new column needs reasoning or adjusted, please.

NotEnsoul:
> Question regarding placement of the new column needs reasoning or adjusted, please.

i was thinking about the case where the system has no gpu available such as a headless server

dmex (approved):

PR #3006, opened 2026-07-22
Remove underlines between link badges

Abrar74774:
Looks better this way imo


Abrar74774:

dmex (approved):

dmex:
Is this the same? [README.md](https://github.com/jxy-s/systeminformer/blob/master/README.md)

Abrar74774:
You mean the underlines showing? yeah same issue. I found the same thing on other repos as well, so possibly a common 'bug'.

rich-ayr:
@dmex Confirming it's not specific to this repo: it's a GitHub-wide markdown rendering issue, so the fork and any README using this pattern will show it too.

With the `<a>` and `` on separate lines, the newline+indent between them is a text node *inside* the anchor, so it inherits the link's text-decoration and renders as an underlined space. Collapsing each pair onto one line removes that text node. That's the minimal fix, and it keeps the newlines between anchors so the badges stay spaced.

_(Markdown badge syntax avoids it for the same reason, but isn't applicable here since the centering needs the HTML wrapper.)_

GitHub could fix it globally with one rule, since flex containers discard whitespace-only children:
```css
.markdown-body a[href]:has(img) { display: inline-flex; }
```
See also https://github.com/orgs/community/discussions/154640

Abrar74774:
This can be merged, right?

PR #3031, opened 2026-08-19
proctree: Fix I/O delta columns sorting by cumulative values

pasharm:
## Description

Sorting the process list by **I/O reads delta**, **I/O writes delta** or **I/O other delta** does not order the rows by the numbers shown in the column. The rows are ordered by each process' cumulative I/O counter since process start instead, which makes the list look unsorted.

Observed on 4.0.26206.2232. Verified by comparing the resulting row order against `Win32_Process.OtherOperationCount`: after clicking **I/O other delta** the order matched the cumulative counter exactly for every visible row, while the displayed per-interval numbers were unordered (108, 512, 4, 22, 4, 161, ...).

### Root Cause

`c4f8951b` ("proctree support aggregates when sorting") converted the sort functions to `PhpAggregateFieldIfNeeded`. For the three I/O delta columns the field offset ended up pointing at `.Value` (the cumulative counter) instead of `.Delta` (the per-interval increment):

```c
BEGIN_SORT_FUNCTION(IoOtherDelta)
{
    ...
    PhpAggregateFieldIfNeeded(node1, ..., FIELD_OFFSET(PH_PROCESS_ITEM, IoOtherCountDelta.Value), &number1);
```

while the cell text for the same column uses `.Delta`:

```c
case PHPRTLC_IOOTHERDELTA:
    PhpAggregateFieldIfNeeded(node, ..., FIELD_OFFSET(PH_PROCESS_ITEM, IoOtherCountDelta.Delta), &value);
```

and so does the header total (`PhpAggregateFieldTotal`, `PHPRTLC_IOOTHERDELTA`).

Before that commit the comparator read `uint64cmp(processItem1->IoOtherCountDelta.Delta, processItem2->IoOtherCountDelta.Delta)`, so 3.2.25011.2103 sorts these columns correctly.

Two consequences worth noting:

* the delta columns currently sort identically to their cumulative counterparts (**I/O reads**, **I/O writes**, **I/O other**), so the two column pairs are indistinguishable when sorted;
* the remaining delta columns are unaffected — `CyclesDelta`, `ContextSwitchesDelta`, `PageFaultsDelta` and `PrivateBytesDelta` already compare `.Delta`. The three I/O ones are the only outliers.

### Fix

Compare `.Delta` in `IoReadsDelta`, `IoWritesDelta` and `IoOtherDelta`, which restores the pre-`c4f8951b` behaviour and matches both the cell text and the header total for those columns.

Aggregation of collapsed child processes is preserved: the change only moves the field offset within the same `PH_UINT64_DELTA`, and the display path already aggregates that exact field.

### Steps to Verify

1. Right-click the process list header → **Choose columns…** and add **I/O other delta**.
2. Click the column header to sort by it, descending.
3. Before the fix: the displayed numbers are in no particular order; the row order instead follows the cumulative "other" operation count (visible in the **I/O other** column, or via `Get-CimInstance Win32_Process | Select ProcessId, OtherOperationCount`).
4. After the fix: the displayed numbers are in descending order, so the top row is the process doing the most "other" I/O operations per update interval.
5. Repeat for **I/O reads delta** and **I/O writes delta**.


pasharm:

dmex (approved):

Issue #2909, opened 2026-05-11
Process-wide Module exports search


ActivistInvestor:
### Description of the feature, modification, idea or suggestion

While it may not be that common, I frequently have a need to search many loaded modules for exports. One of the processes that I routinely work with has at least 200 modules loaded, whose exports must be searched for a specific API. Perhaps it could be handled with an extension of some sort.

Issue #2813, opened 2026-02-03
[Feature request] Event-driven process creation tracking (ETW / kernel notify) to avoid missing short-lived processes


vandir:
## Problem

System Informer (like Process Hacker) currently discovers processes by periodically enumerating them using a polling mechanism.

This polling/snapshot approach inherently misses processes that are created and terminated between two refresh intervals.

In malware analysis and incident response scenarios, this is a significant limitation because many modern samples deliberately spawn very short-lived helper processes (living for a few milliseconds) specifically to evade user-mode monitoring tools based on polling.

Even with the minimum refresh interval, these transient processes are never observed by System Informer.

## Expected behavior

System Informer should be able to observe **all** process creation and termination events, including processes that exist only for a very short time.

This is already technically possible on Windows using:

- `PsSetCreateProcessNotifyRoutineEx` (kernel callback)
- or, from user mode, the ETW provider `Microsoft-Windows-Kernel-Process`

Both mechanisms are event-driven and originate from the same kernel notification path (`PspCallProcessNotifyRoutines`), meaning no process can be missed, regardless of lifetime.

Tools like:
- Sysmon
- Procmon
- WPR/WPA

are able to capture these processes reliably because they rely on event notifications rather than snapshots.

## Proposed feature

Add an optional **event-driven process tracking mode** based on ETW `Microsoft-Windows-Kernel-Process` provider, used to:

- Track process start/stop events in real time
- Populate the process list and history from ETW events instead of relying only on periodic snapshots
- Maintain a history/log of processes that existed too briefly to be caught by polling

This could be implemented as:

- An optional ETW session started by System Informer
- A background listener feeding process creation/termination events into the UI
- A "transient processes" history view

## Why this matters

This is particularly relevant for:

- Malware analysis
- EDR research
- IR / forensic workflows
- Debugging software that spawns short-lived helper processes

Currently, System Informer systematically misses this entire class of processes.

## Additional notes

This is not a performance/refresh interval issue but an architectural limitation of snapshot-based enumeration.

Using ETW would align System Informer with how modern EDRs and system tracing tools observe process activity.

## Technical references

Use the ETW kernel process provider:

- Provider: Microsoft-Windows-Kernel-Process  
- GUID: {22FB2CD6-0E7B-4E8F-9D10-4A8A7C0C0F2F}  
- Event ID 1 → Process Start  
- Event ID 2 → Process Stop  

User-mode APIs for real-time consumption:

[...]

Issue #2649, opened 2025-08-09
add "Startup apps" tab


henrypp:
### Description of the feature, modification, idea or suggestion

add "Startup apps" tab same as in  **taskmgr.exe**

<img width="248" height="578" alt="Image" src="https://github.com/user-attachments/assets/6592d169-39a8-444f-817f-7d5fb9366566" />

i can create pull request for it, but I haven't decided yet create tab "Startup apps" or create plugin for it, need your decide

dmex:
https://github.com/winsiderss/systeminformer/issues/2442#issuecomment-2784839879

henrypp:
> The first issue would be how Windows Task Manager is using undocumented interfaces for application startup entries:

there is no requirement for use undocumented interfaces, i think registry&startup folder enum is enough:
```
Software\\Microsoft\\Windows\\CurrentVersion\\Run
Software\\Microsoft\\Windows\\CurrentVersion\\RunOnce
Software\\Microsoft\\Windows\\CurrentVersion\\RunOnceEx
Software\\Microsoft\\Windows\\CurrentVersion\\RunServices (?)
Software\\Microsoft\\Windows\\CurrentVersion\\RunServicesOnce (?)
%APPDATA%\\Microsoft\\Windows\\Start Menu\\Programs\\Startup
```
undocumented interfaces is for disable autoruns items, i dont know for what this interfaces?

> there might be other issues considering Sysinternals Autoruns also doesn't show these entries.

wat entries Autoruns dont show?

henrypp:
how it looks like you can see [here](https://github.com/henrypp/autoruns2?tab=readme-ov-file#readme), i only need your advice

PR #3035, opened 2026-08-22
More Features 

StathisDC:
Exports search (#2909) A "Search exports" option in the Modules tab. It scans every DLL a process has loaded and lists all exported functions with their module, ordinal and address, so you can find which module provides a given API without checking each one manually

Process Tracker (#2813) A window under Tools > System that uses ETW to log process creation and termination as they happen. Since it's event-driven rather than polled, it captures short-lived processes that the regular process list misses.

Startup Apps (#2649) A startup manager under Tools > System, similar to Task Manager's Startup tab. Shows entries from the Run registry keys and startup folders, indicates enabled/disabled state, and lets you enable, disable or delete them.


StathisDC:

